PR #127, opened 2023-05-08
Add System Parachain Asset Ids

marshacb:
* update create registry script to include Statemint and Statemine asset ids in registry

### DIFF
--- a/src/registry/types.ts
+++ b/src/registry/types.ts
@@ -4,7 +4,7 @@ export type ChainInfo = {
 	[x: string]: {
 		specName: string;
 		tokens: string[];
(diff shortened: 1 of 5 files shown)


src/registry/types.ts
```diff
@@ -4,6 +4,7 @@ export type ChainInfo = {
 	[x: string]: {
 		specName: string;
 		tokens: string[];
+		assetIds?: number[];
```
TarikGul, line 7: I think this can just be `assetIds: number[]` since if its not a common good parachain it will just be empty

TarikGul (approved):

Issue #128, opened 2023-05-09
Check that Integer AssetId inputs are Valid System Parachain Asset Ids


marshacb:
For XCM's of system parachain origin we should check that string number asset id inputs are valid and exist

PR #129, opened 2023-05-09
check integer asset ids are valid system

marshacb:
* adds check to ensure integer asset id inputs exist on each respective system parachain
* adds  test for system to para
* adds test for para to para

closes [#128](https://github.com/paritytech/asset-transfer-api/issues/128)

### DIFF
--- a/src/errors/checkXcmTxInputs.ts
+++ b/src/errors/checkXcmTxInputs.ts
@@ -16,14 +16,40 @@ export const checkAssetIdInput = (
 	assetIds: string[],
 	relayChainInfo: ChainInfo,
 	specName: string,
-	destChainId: string
+	destChainId: string,
+	xcmDirection: IDirection
 ) => {
 	for (let i = 0; i < assetIds.length; i++) {
 		const assetId = assetIds[i];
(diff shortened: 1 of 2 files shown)


src/errors/checkXcmTxInputs.ts
```diff
@@ -16,14 +16,40 @@ export const checkAssetIdInput = (
...
 	for (let i = 0; i < assetIds.length; i++) {
 		const assetId = assetIds[i];
 		const parsedAssetIdAsNumber = Number.parseInt(assetId);
-		const notValidNumber = Number.isNaN(parsedAssetIdAsNumber);
+		const invalidNumber = Number.isNaN(parsedAssetIdAsNumber);
+
+		// check that the specific assetId exists for the destChain
+		if (!invalidNumber) {
+			if (xcmDirection.toLowerCase().startsWith('system')) {
+				const systemParachainId = 1000;
```
TarikGul, line 30: We have a default ID in `consts.ts` called `SYSTEM_PARACHAINS_IDS` with 1 value in it for the id. You may want to use that as it would make everything come from one place.

TarikGul (approved):
1 suggestion, but looks great to me!

PR #130, opened 2023-05-10
Remove registry code, and import registry from asset-transfer-api-registry

TarikGul:
This removes any and all code that has to do with the registry, and instead imports the registry from `@substrate/asset-transfer-api-registry`


marshacb (approved):
awesome. lgtm!

Issue #132, opened 2023-05-11
Re standardize our type naming be removing the `I` prefix for interfaces


TarikGul:
Lets remove the `I` in front of any interface type and just have the names be there normal names.

PR #133, opened 2023-05-11
Refactor Type and Interface Declarations

Rymul:
This PR standardizes the naming convention for type and interface declarations to remove the I and use standard names.

closes: https://github.com/paritytech/asset-transfer-api/issues/132


TarikGul (approved):
Thanks for the contribution! 👍

PR #135, opened 2023-05-12
Add teleport logic to AssetsTransferApi

TarikGul:
Waiting on https://github.com/paritytech/asset-transfer-api/issues/136 to be resolved for `SystemToPara` but I will continue to work on the rest of the logic.

### DIFF
--- a/src/errors/checkXcmTxInputs.ts
+++ b/src/errors/checkXcmTxInputs.ts
@@ -54,16 +54,6 @@ export const checkAssetIdInput = (
 			let isValidTokenSymbol = false;
 			const chainInfo = relayChainInfo[destChainId];
 
-			const isNativeChain =
-				chainInfo.specName.toLowerCase() === specName.toLowerCase();
-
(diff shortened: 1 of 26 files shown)


src/errors/checkXcmTxInputs.ts
```diff
@@ -57,6 +57,7 @@ export const checkAssetIdInput = (
 			const isNativeChain =
 				chainInfo.specName.toLowerCase() === specName.toLowerCase();
 
+			// TODO: This logic breaks when we have 2 chains that share the same native tokens.
```
TarikGul, line 60: @marshacb So this logic below has an edge case we didn't cover. Such as what if the 2 chains are statemine, and polkadot. It will fail since they both share native tokens. 

You can see the failing tests in the test suite for github actions.

TarikGul: Hey so I removed it, as I think the check might be a little redundant. Let me know what you think. We can add it back in when we do specific checks on direction.

marshacb (approved):
lgtm!

PR #137, opened 2023-05-12
add fee estimate

marshacb:
* adds method to calculate fee estimate for a given extrinsic

closes: [issue #32](https://github.com/paritytech/asset-transfer-api/issues/32)


src/AssetsTransferApi.ts
```diff
@@ -156,6 +160,44 @@ export class AssetsTransferApi {
 			opts?.format
 		);
 	}
+	/**
+	 * Fetch estimated fee information for an extrinsic
+	 *
+	 * @param tx TxResult<T>
```
TarikGul, line 166: `@param format` as well.

src/AssetsTransferApi.ts
```diff
@@ -156,6 +160,44 @@ export class AssetsTransferApi {
...
+	 */
+	public async fetchFeeInfo<T extends Format>(
+		tx: SubmittableExtrinsic<'promise', ISubmittableResult>,
+		format?: T
+	): Promise<RuntimeDispatchInfo | RuntimeDispatchInfoV1 | null> {
+		const { _api: api } = this;
```
TarikGul, line 172: Should we keep this in line with the rest of the API where its `_api` to denote its read only, if you think its easier to read like `api` then I am happy to have a follow up PR change the rest of the API.

TarikGul (approved):
lgtm, great work :)

src/AssetsTransferApi.ts
marshacb: totally. Ill change this to `_api`

PR #142, opened 2023-05-25
Fix parents for native vs non-native tokens when construct xcm types for assets

TarikGul:
This ensures when we create xcm types for assets for `SystemToPara` we correctly input the right `parents` value depending on the token.


marshacb (approved):
lgtm!

PR #143, opened 2023-05-26
Organize Integration tests, and add V3 tests for all Xcm Directions

TarikGul:
This PR focuses on:

- Adding V3 integration tests for all xcm directions
- Refactoring some of the tests for `AssetsTransferApi` so that they are properly organized and not mixed in with unit testing.


marshacb (approved):
looks good!

PR #144, opened 2023-05-26
Reorganize zombienet, and add small network config

TarikGul:
This PR adds a new structure to the zombienet networks. I added a small, and medium network depending on the test cases.


marshacb (approved):
nice!

Issue #160, opened 2023-05-31
Decoded extrinsics functionality


TarikGul:
It would be nice to add a public function to the `AssetsTransferApi` that will decode a constructed `call`, `payload`, or `submittableExtrinsic`.

PR #167, opened 2023-06-01
add ability to decode an extrinsic

marshacb:
* adds a method to the AssetTransferApi to decode an extrinsic given the hex of its tx 

closes: [#160](https://github.com/paritytech/asset-transfer-api/issues/160)


src/AssetsTransferApi.ts
```diff
@@ -397,4 +397,44 @@ export class AssetsTransferApi {
...
+		if (fmt === 'payload') {
+			const extrinsicPayload = _api.registry.createType(
+				'ExtrinsicPayload',
+				encodedTransaction,
+				{
+					version: 4,
+				}
+			);
+
+			const extrinsicMethodInfo = extrinsicPayload.method.toHuman() as string;
```
TarikGul, line 422: Why do we do `toHuman()` here but `toString` in the other two formats? Any way to keep them all the same?

TarikGul:
I just added this issue: https://github.com/paritytech/asset-transfer-api/issues/166 and it kind of relates to this PR, Since you added the westmint mock api. We should update the other two metadata's to be 9420 at the least. It might require some fixing of the tests results which is a bit of a pain, but it would probably do us good in the future.

TarikGul (approved):
Some nits, and questions but overall, looks great!

src/testHelpers/mockWestmintApi.ts
```diff
@@ -0,0 +1,4 @@
+import { createApiWithAugmentations } from './createApiWithAugmentations';
```
TarikGul, line 1: Copyright

src/AssetsTransferApi.ts
marshacb: the tx method field for the extrinsic payload is of type `Bytes` so I called `toHuman` in order to first turn then into an object. We can call toHuman()?.toString() but ultimately we can't directly `toString` from `Bytes` since we essentially get the hex and not the human readable object. Ill update this to call `toString` after in case thats more clear.

Issue #47, opened 2023-03-14
Implement `paysWithFeeDest` option


TarikGul:

TarikGul:
rel: https://github.com/paritytech/asset-transfer-api/issues/32

`feeAssetItem` resolves to index 0 for all versions under or equal to V2.

PR #176, opened 2023-06-06
implement pays with fee dest

marshacb:
* if `paysWithFeeDest` option is provided for `reserveTransferAssets` or `limitedReserveTransferAssets` XCM V3 txs, correctly finds and assigns the index of the desired fee asset to the XCMs feeAssetItem
*  errors if given an invalid asset
*  defaults to 0 if no match is found
* typedoc doc updates

closes: [#47](https://github.com/paritytech/asset-transfer-api/issues/47)


src/util/applyPaysWithFeesDest.ts
```diff
@@ -0,0 +1,87 @@
...
+export const applyPaysWithFeeDestination = (
+	paysWithFeeDest: string,
+	multiAssets: MultiAsset[],
+	specName: string
+): MultiAsset | string => {
+	let result: MultiAsset | string =
+		'destination chain fee asset was not updated';
+
+	if (paysWithFeeDest) {
+		console.log('what is paysWithFeesDest', paysWithFeeDest.toLowerCase());
```
TarikGul, line 33: We should remove this console.

src/util/applyPaysWithFeesDest.ts
```diff
@@ -0,0 +1,87 @@
...
+		const isRelayFeeAsset =
+			paysWithFeeDest.toLowerCase() === 'dot' ||
+			paysWithFeeDest.toLowerCase() === 'ksm' ||
+			paysWithFeeDest.toLowerCase() === 'wnd' ||
+			paysWithFeeDest.toLowerCase() === '0';
+
+		console.log('isRelayAsset', isRelayFeeAsset);
```
TarikGul, line 40: Same with this console.

TarikGul:
> if paysWithFeeDest is provided for XCM V3 txs, correctly finds and assigns the desired fee MultiAsset to index 0

So I think here what we do is assign the asset index inside of the multiasset to the `feeAssetItem` field for the xcm trasnfer which is going to be a u32. For example: https://github.com/polkadot-js/api/blob/master/packages/types-augment/src/lookup/types-polkadot.ts#L952-L959

I think if we apply this logic it will break because we will reorder the already ordered assets. So after we sort it I believe we need to find the index, and assign that index to `feeAssetItem`.

cc: @IkerAlus

IkerAlus:
There are two points that we should take care and not mix unnecessarily:

- Reordering assets beforehand so the array input in `assets` is in the right order.
- Deriving the right index (`u32`) for `feeAssetItem`  to specify the asset paying the fee as the position in the `assets` vector above.

As far as I can see, the user is asked to provide an assetId as the `paysWithFeeDest` input and not an integer right? In that case, then the exposed API function can implement all the reordering logic and error handling internally, this is, first reorder the assets and create the vector fro `assets` input and then derive the right integer for `feeAssetItem`.

I recommend you to read this issues to get a better context on how this should be dealt with:

- https://github.com/paritytech/polkadot/issues/6074
- https://github.com/polkadot-js/api/issues/4537

src/createXcmCalls/reserveTransferAssets.ts
```diff
@@ -43,5 +46,21 @@ export const reserveTransferAssets = (
 		assetIds
 	);
 
-	return ext(dest, beneficiary, assets, 0);
+	let feeAssetItem = 0;
```
TarikGul, line 49: Lets apply this logic to `createFeeAssetItem` for `ICreateXcmTypes` for each direction

src/createXcmTypes/util/createSystemToParaMultiAssets.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Creates and returns a MultiAsset array for system parachains based on provided specName, assets and amounts 
+ *
+ * @param api ApiPromise[]
+ * @param amounts string[]
+ * @param specName string
+ * @param assets string[]
+ */
+export const createSystemToParaMultiAssets = (
```
TarikGul, line 21: We can add this directly inside of the SystemToPara file as a seperate helper function.

src/util/getFeeAssetItemIndex.ts
```diff
@@ -0,0 +1,69 @@
...
+	specName: string
+): number => {
+	let result = 0;
+
+	if (paysWithFeeDest) {
+		const isRelayFeeAsset =
+			paysWithFeeDest.toLowerCase() === 'dot' ||
+			paysWithFeeDest.toLowerCase() === 'ksm' ||
+			paysWithFeeDest.toLowerCase() === 'wnd' ||
+			paysWithFeeDest.toLowerCase() === '0';
```
TarikGul, line 27: Doesn't need to be `toLowerCase` for `0`

TarikGul (approved):

PR #178, opened 2023-06-07
Check base types for createTransferTransaction inputs

TarikGul:
If the library is used in plain JS, we wont have the same type verification for inputs, therefore we need to enforce the types but adding our own base checks.


marshacb (approved):
looks good!

PR #179, opened 2023-06-07
Rework AssetTransferApi constructor, and constructApiPromise

TarikGul:
### `AssetTransferApi`

`AssetTransferApi` now takes in 2 arguments for the constructor instead of 2:

```
api: ApiPromise,
specName: string,
safeXcmVersion: number,
opts: AssetsTransferApiOpts = {}
```

### `constructApiPromise`

`constructApiPromise` now returns an object with all the necessary args for AssetTransferApi:

```typescript
	return {
		api,
		specName: specName.toString(),
		safeXcmVersion: safeXcmVersion.toNumber(),
	};
```


marshacb (approved):
other than lint fixes, looks great!

PR #183, opened 2023-06-08
Rework the Registry to be a public class

TarikGul:
## Summary

I reworked the registry to save on performance, and memory while also exposing a bunch of public functions to the user. 

### Exposed functions from the registry

```typescript
// Getter for the complete registry
public get getRegistry

// Getter for the name of the relay chain for this network
public get getRelayChain

// Getter for the registry associated with this networks relay chain
public get getRelaysRegistry

// Lookup all chains that have the following token symbol. It will return an array
// with all the chains that have the following token symbols. Note this will only
// be searched in the respective relay chains registry.
public lookupTokenSymbol(symbol: string)

// Lookup all chains that have the following assetId. It will return an array
// with all the chains that have the following AssetId. Note this will only
// be searched in the respective relay chains registry.
public lookupAssetId(id: string)

// Check whether a parachain id exists within the relay chains registry.
public lookupParachainId(id: string)

// Return the info for a parachain within a relay chains registry.
public lookupParachainInfo(id: string)
```

### DIFF
--- a/src/AssetsTransferApi.spec.ts
+++ b/src/AssetsTransferApi.spec.ts
@@ -90,7 +90,6 @@ describe('AssetTransferAPI', () => {
 		describe('SystemToRelay', () => {
 			it('Should corectly return Native', () => {
 				const assetType = systemAssetsApi['fetchAssetType'](
-					'statemint',
 					'0',
 					['DOT'],
 					Direction.SystemToRelay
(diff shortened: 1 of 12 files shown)


marshacb (approved):

marshacb (commented):
lgtm!

Issue #140, opened 2023-05-25
Add MultiLocation Support for AssetIds


marshacb:
1. AssetId validation
2. System to Para MultiLocation Construction

PR #189, opened 2023-06-13
Fix issue with SystemToPara call types

TarikGul:
closes: https://github.com/paritytech/asset-transfer-api/issues/188
rel: https://github.com/paritytech/asset-transfer-api/issues/140

This ensures that we send assets, and native tokens from system parachains as reserve transfers to parachains. When `MultiLocation` support active for trusted tokens we can adjust this logic to take that into account.


marshacb (approved):
looks good!

Issue #165, opened 2023-06-01
Add examples directory 


TarikGul:
This should work through as many examples as possible.

- Work through each xcmdirection
    - [x] RelayToPara
    - [x] RelayToSystem
    - [x] SystemToPara
    - [x] SystemToRelay
- Each XcmDirection should also showcase working through each format.
- We should showcase using the options as well, in the examples
- There should be an example with a submittableExtrinsic and signing then sending. So kind of an e2e example of the workflow.

PR #193, opened 2023-06-14
Add inital examples template

TarikGul:
rel: https://github.com/paritytech/asset-transfer-api/issues/165

This is the inital examples template. We will expand on this and add more depth, refer to the issue for more context


examples/relayToSystem.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
```
IkerAlus, line 5: can we add as a comment a short description of the example? For example, in this case it would be something like:

"In this example we are creating a call to send X amount of WND from Westend relaychain to a user account in parachain Y and with a limit in execution weight at the destination".

IkerAlus (changes requested):
Consider the comment to add a small description

examples/relayToSystem.ts
TarikGul: Updated

IkerAlus (commented):
LGTM!

Issue #190, opened 2023-06-13
Throw an error when an asset symbol has multiples in the registry


TarikGul:
Currently we allow users to input both assetIds and symbols for their transaction. But we need to ensure that the asset symbol does not have duplicates in the registry, and if it does we need to throw an error.

The error should be very descriptive, and return all the assets ids that have the same asset symbol.

PR #195, opened 2023-06-14
error on multiple registry entries

marshacb:
* throws an error which includes the assetIds/symbols of matched assets when multiple entries are found in a chains asset registry for a given symbol
* adds unit tests

closes: [#190](https://github.com/paritytech/asset-transfer-api/issues/190)


src/errors/checkXcmTxInputs.ts
```diff
@@ -202,16 +202,44 @@ const checkSystemToParaAssetId = (
 				}
 			}
 
+			type AssetInfo = {
```
TarikGul, line 205: We should declare this at the top level.

TarikGul (approved):
lgtm, just one small nit

PR #198, opened 2023-06-15
Rethread Registry down to createAssets, and createFeeAssetItem as an option

TarikGul:
This rethreads the `registry` as an `opt` down to `SystemToPara` in order to avoid calling `parseRegistry`.

The layers of threading are:

`AssetTransferApi` -> `extrinsic call` -> `createXcmTypes`


marshacb (approved):
looks good!

PR #215, opened 2023-06-27
feat:`SystemToSystem`

marshacb:
- [x] SystemToSystem examples
- [x] SystemToSystem error layer
- [x] integration tests for SystemToSystem
- [x] tests for SystemToSystem

Note: Requires XCM config updates for each System chain to regard one another as trusted teleport locations.

closes: [#210](https://github.com/paritytech/asset-transfer-api/issues/210)


src/AssetsTransferApi.ts
```diff
@@ -186,7 +227,12 @@ export class AssetsTransferApi {
 
 		let txMethod: Methods;
 		let transaction: SubmittableExtrinsic<'promise', ISubmittableResult>;
-		if (assetType === AssetType.Foreign) {
+		if (
+			assetType === AssetType.Foreign ||
+			(assetType === AssetType.Native &&
```
TarikGul, line 232: Can we abstract this second conditional after the `||` and assign it to a variable. I think it would make it a lot more readable.

src/AssetsTransferApi.ts
```diff
@@ -329,6 +375,10 @@ export class AssetsTransferApi {
 			return Direction.SystemToRelay;
 		}
 
+		if (isSystemParachain && SYSTEM_PARACHAINS_IDS.includes(destChainId)) {
```
TarikGul, line 378: I saw a comment by @joepetrowski  that all `para_ids` less than 2000 can be considered a system parachain, so that might help us from having to have an array with all the ID's to identify if its a system para. 

I think in this PR we can keep it as is, but its good to keep that in mind. I'll write an issue to keep track of this for us.

src/AssetsTransferApi.ts
```diff
@@ -524,7 +581,11 @@ export class AssetsTransferApi {
 		const submittableData: SubmittableMethodData = JSON.parse(
 			submittableString
 		) as unknown as SubmittableMethodData;
-		const addr = submittableData.method.args.dest.id;
+
+		const addr: string = submittableData.method.args.target
```
TarikGul, line 585: Hmm what will happen here if the string is empty? Will it ever reach that?

src/consts.ts
```diff
@@ -12,11 +12,19 @@ export const RELAY_CHAIN_IDS = ['0'];
 /**
  * List of all known system parachains.
  */
-export const SYSTEM_PARACHAINS_NAMES = ['statemine', 'statemint', 'westmint'];
+export const SYSTEM_PARACHAINS_NAMES = [
```
TarikGul, line 15: Lets add `asset-hub-polkadot`, `asset-hub-kusama`, and `asset-hub-westend` as well.

src/AssetsTransferApi.ts
marshacb: good catch! Ill do that and assign it to a variable :)

src/AssetsTransferApi.ts
marshacb: exact same thought came to my mind when I saw that. Agree we should keep it in mind so an issue is great

src/createXcmTypes/SystemToSystem.ts
```diff
@@ -0,0 +1,302 @@
...
+	 * @param accountId The accountId of the beneficiary
+	 * @param xcmVersion The accepted xcm version
+	 */
+	createBeneficiary: (
+		api: ApiPromise,
+		accountId: string,
+		xcmVersion?: number
+	): VersionedMultiLocation => {
+		if (xcmVersion == 2) {
+			const X1 = isEthereumAddress(accountId)
```
TarikGul: I dont think we need the `isEthereum` here, and we can get rid of `AccountId20` since the system parachain probably wont support ethereum addresses.

src/createXcmTypes/SystemToSystem.ts
```diff
@@ -0,0 +1,302 @@
...
+				V2: {
+					parents: 0,
+					interior: {
+						X1,
+					},
+				},
+			});
+		}
+
+		const X1 = isEthereumAddress(accountId)
```
TarikGul: Same as above

src/types.ts
```diff
@@ -189,7 +193,7 @@ export interface ChainInfo {
 }
 
 export type NonRelayNativeInterior = {
-	X2: [{ PalletInstance: string }, { GeneralIndex: string }];
+	X2: [{ PalletInstance: string | undefined }, { GeneralIndex: string }];
```
TarikGul, line 196: When will the palletInsance be `undefined`, i might have missed it in the above.

TarikGul (approved):
Overall, amazing PR. There were a ton of things to tackle so this was really well done. Just a couple of nits and questions before we merge.

src/types.ts
marshacb: If I recall the assets pallet doesnt exist for some chains such as collectives (https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/collectives/collectives-polkadot/src/lib.rs) and we do throw an error for cases where its not found.

However, I think this is more a result of how I placed the palletId in SystemToSystem so Im going to refactor this back to be a string only.

TarikGul: Awesome sounds good :) Thanks for the explanation

src/AssetsTransferApi.ts
marshacb: great catch! I'll update this to cover the case for local transactions as well as V2 and V3 MultiAssets

Issue #106, opened 2023-04-12
Testnetwork needs to be adjusted to work with V2 and V3


TarikGul:

TarikGul:
So currently we need to get the local testnetwork working. This means that we should be able to run `medium-network.toml` with zombienet, then run `yarn start:zombienet-post-script` which should succesfully create a test asset on our local system parachain. 

The current issue is when the xcm transfer is sent that there is a `Barrier` xcm error on the system parachain end.

@bee344 I would suggest reading the zombienet section of the README.md, and getting the network working locally. Then try running the script and monitoring the messages sent through polkadot-js.

PR #223, opened 2023-07-12
fix: added unpaidExecution for testnet XCM message

bee344:
I modified the XCM Message to start with `unpaidExecution` so that it passes the barrier specified by `xcm_config.rs` on Polkadot and Kusama. Solves issue #106 .


TarikGul (approved):
Really amazing! Great Job. Just need to `yarn lint:fix`

PR #224, opened 2023-07-13
feat: add support for foreign asset multilocations

marshacb:
TODOs: 
- [x] getFeeAssetItemIndex for foreign assets
- [x] update multiasset sorting
- [x] integration tests
- [x] unit tests
- [x] docs
- [x] update examples
- [x] error validation layer
- [x] add foreign assets pallet instance from registry

closes: [#140](https://github.com/paritytech/asset-transfer-api/issues/140)
closes: [#229](https://github.com/paritytech/asset-transfer-api/issues/229)
closes: [#230](https://github.com/paritytech/asset-transfer-api/issues/230)


src/AssetsTransferApi.spec.ts
```diff
@@ -478,4 +555,38 @@ describe('AssetTransferAPI', () => {
 			);
 		});
 	});
+
+	describe('multiLocationAssetIsDestParachainsNativeAsset', () => {
+		it('Should correctly return true when a foreign assets multilocation matches its native chain of origin', () => {
+			const expected = true;
```
TarikGul, line 561: I think we can just do `expect(isNativeChain).toEqual(true);` instead of saving this to a variable.

src/AssetsTransferApi.spec.ts
```diff
@@ -478,4 +555,38 @@ describe('AssetTransferAPI', () => {
...
+				'{"parents":"1","interior":{"X1": {"Parachain":"2023"}}}';
+			const destChainId = '2023';
+
+			const isNativeChain =
+				systemAssetsApi.multiLocationAssetIsDestParachainsNativeAsset(
+					destChainId,
+					multiLocationAssetId
+				);
+
+			expect(isNativeChain).toEqual(expected);
```
TarikGul, line 573: For this one expect(isNativeChain).toEqual(false); instead of saving this to a variable. Similar to the above.

src/AssetsTransferApi.spec.ts
```diff
@@ -478,4 +555,38 @@ describe('AssetTransferAPI', () => {
...
+		});
+
+		it('Should correctly return false when a foreign assets multilocation does not match the destination chain', () => {
+			const expected = false;
```
TarikGul, line 577: Same with this one.

src/AssetsTransferApi.ts
```diff
@@ -234,10 +265,39 @@ export class AssetsTransferApi {
 			xcmDirection === Direction.SystemToSystem &&
 			!containsNativeRelayAsset(assetIds, nativeRelayChainAsset);
 
-		if (assetType === AssetType.Foreign || isSystemToSystemReserveTransfer) {
+		// if transfering a foreign asset, we determine if the asset is the native asset
+		// of the dest parachain, if so we construct a teleport
+		const isSystemToParaTeleport =
```
TarikGul, line 270: What do you think if we put all this logic in `isSystemToParaTeleport`, and `isSystemToParaReserveTransfer` inside of `checkAssetType`. Then we can change the return `enum` to be something like 

```
export enum AssetCallType {
	Reserve = 'Reserve',
	Teleport = 'Teleport',
}
```

That we things can be a little more abstracted, and have a single source of truth.

TarikGul: Then the If statement should be as simple as `if (assetCallType === AssetType.Reserve)`

src/createXcmCalls/limitedTeleportAssets.ts
```diff
@@ -31,26 +31,32 @@ export const limitedTeleportAssets = (
...
-	const assets = typeCreator.createAssets(
+	const assets = await typeCreator.createAssets(
 		api,
 		normalizeArrToStr(amounts),
 		xcmVersion,
 		specName,
 		assetIds,
-		{ registry }
+		{ registry },
+		transferForeignAssets
```
TarikGul, line 50: 1. Can we name `transferForeignAssets` to `isForeignAssetsTransfer`. This would allow things to make sense with the boolean. 

2. Then `isForeignAssetsTransfer` should be under the options field along with the registry. This is because when we add liquid tokens, there will be a field as well that says `isLiquidToken` transfer, and it will allow us to organize our boolean logic for which kind of transfer it should be.

src/createXcmCalls/limitedTeleportAssets.ts
```diff
@@ -31,26 +31,32 @@ export const limitedTeleportAssets = (
 	xcmVersion: number,
 	specName: string,
 	registry: Registry,
+	transferForeignAssets: boolean | undefined,
```
TarikGul, line 34: lets do `transferForeignAssets?: boolean` instead.

src/createXcmCalls/reserveTransferAssets.ts
```diff
@@ -31,31 +31,37 @@ export const reserveTransferAssets = (
 	xcmVersion: number,
 	specName: string,
 	registry: Registry,
+	transferForeignAssets: boolean | undefined,
```
TarikGul, line 34: transferForeignAssets?: boolean

src/createXcmCalls/reserveTransferAssets.ts
```diff
@@ -31,31 +31,37 @@ export const reserveTransferAssets = (
...
+				api,
+				{
+					registry,
+					paysWithFeeDest,
+					specName,
+					assetIds,
+					amounts,
+					xcmVersion,
+				},
+				transferForeignAssets
```
TarikGul, line 63: This should be added into the options above.

src/createXcmCalls/teleportAssets.ts
```diff
@@ -31,24 +31,30 @@ export const teleportAssets = (
 	xcmVersion: number,
 	specName: string,
 	registry: Registry,
+	transferForeignAssets: boolean | undefined,
```
TarikGul, line 34: `transferForeignAssets?: boolean`

src/createXcmTypes/ParaToSystem.ts
```diff
@@ -102,19 +103,22 @@ export const ParaToSystem: ICreateXcmType = {
...
 		api: ApiPromise,
 		amounts: string[],
 		xcmVersion: number,
 		specName: string,
 		assets: string[],
-		opts: CreateAssetsOpts
-	): VersionedMultiAssets => {
-		const sortedAndDedupedMultiAssets = createParaToSystemMultiAssets(
+		opts: CreateAssetsOpts,
+		transferForeignAssets: boolean | undefined
```
TarikGul, line 113: 1. `transferForeignAssets?: boolean`

In addition to making this part of the options.

src/createXcmTypes/ParaToSystem.ts
```diff
@@ -209,12 +225,14 @@ export const ParaToSystem: ICreateXcmType = {
...
-const createParaToSystemMultiAssets = (
+const createParaToSystemMultiAssets = async (
+	api: ApiPromise,
 	amounts: string[],
 	specName: string,
 	assets: string[],
-	registry: Registry
-): MultiAsset[] => {
+	registry: Registry,
+	transferForeignAssets: boolean | undefined
```
TarikGul, line 234: `transferForeignAssets?: boolean`

src/createXcmTypes/util/getSystemChainAssetId.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Returns the correct asset id for a valid system chain token symbol
+ * integer id or foreign asset multilocation
+ * errors if given an invalid symbol, integer id or multilocation that is
+ * not registered or found in the chains state
+ *
+ * @param tokenSymbol string
+ * @param specName string
+ */
+export const getSystemChainAssetId = async (
```
TarikGul, line 20: What do we think about making the `_api` the first arg here. I think it would be nice if we try and make the api come first wherever applicable as it will help with consistency.

src/createXcmTypes/ParaToSystem.ts
```diff
@@ -232,22 +250,31 @@ const createParaToSystemMultiAssets = (
...
+				assetId,
+				specName,
+				api,
+				transferForeignAssets
+			);
 		}
 
-		const interior: MultiAssetInterior = isHex(assetId)
-			? { X2: [{ GeneralKey: assetId }] }
+		const interior: InteriorMultiLocation = isHex(assetId)
```
TarikGul, line 261: Great improvement 👍

src/createXcmTypes/types.ts
```diff
@@ -41,10 +41,15 @@ export interface ICreateXcmType {
 		xcmVersion: number,
 		specName: string,
 		assets: string[],
-		opts: CreateAssetsOpts
-	) => VersionedMultiAssets;
+		opts: CreateAssetsOpts,
+		transferForeignAssets?: boolean | undefined
```
TarikGul, line 45: As mentioned earlier it would be nice to see `transferForeignAssets` to `opts`, and rename it to `isForeignAssetsTransfer`

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright 2023 Parity Technologies (UK) Ltd.
+
+import { ApiPromise } from '@polkadot/api';
+
+import { BaseError } from '../../errors';
+import { Registry } from '../../registry';
+import { ForeignAssetsInfo } from '../../registry/types';
+// import { ForeignAssetMultiLocation } from "src/types";
```
TarikGul, line 8: Commented import

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const foreignAssetMultiLocationIsInRegistry = (
```
TarikGul, line 10: can we make the api the first arg here as well.

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
```diff
@@ -0,0 +1,59 @@
...
+					return true;
+				}
+			}
+		}
+	} catch (error) {
+		const errorInfo = (error as Error).message.split('::');
+		const errorDetails = errorInfo[errorInfo.length - 2].concat(
+			errorInfo[errorInfo.length - 1]
+		);
+		throw new BaseError(`error creating MultiLocation type:${errorDetails}`);
```
TarikGul, line 55:
```suggestion
		throw new BaseError(`Error creating MultiLocation type:${errorDetails}`);
```

src/createXcmTypes/util/getSystemChainAssetId.ts
```diff
@@ -0,0 +1,99 @@
...
+	const newRegistry = new Registry(specName, {});
+	const systemChainId = getChainIdBySpecName(newRegistry, specName);
+
+	if (!SYSTEM_PARACHAINS_IDS.includes(systemChainId)) {
+		throw new BaseError(
+			`specName ${specName} did not match a valid system chain ID. Found ID ${systemChainId}`
+		);
+	}
+
+	if (transferForeignAssets && systemChainId === '1000') {
```
TarikGul, line 36: Should we save `systemChainId === '1000'` to something like `const is assetHubId = systemChainId === '1000'`

src/createXcmTypes/util/getSystemChainAssetId.ts
```diff
@@ -0,0 +1,99 @@
...
+		// determine if we
```
TarikGul, line 37: Unfinished comment

src/createXcmTypes/util/getSystemChainAssetId.ts
```diff
@@ -0,0 +1,99 @@
...
+		const multiLocationIsInRegistry = foreignAssetMultiLocationIsInRegistry(
+			asset,
+			newRegistry,
+			_api
+		);
+
+		if (multiLocationIsInRegistry) {
+			assetId = asset;
```
TarikGul, line 45: Does this assume that the assetId is a `multiLocation` as a string?

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
```diff
@@ -0,0 +1,59 @@
...
+// import { ForeignAssetMultiLocation } from "src/types";
+
+export const foreignAssetMultiLocationIsInRegistry = (
+	multilocationStr: string,
+	registry: Registry,
+	api: ApiPromise
+): boolean => {
+	try {
+		const assetHubChainId = 1000;
+		const multiLocation = api.registry.createType(
```
TarikGul, line 17: Would this have to be in a try-catch since it will throw an error if string is incorrect, or do we do error validation on this inside of checkXcmTxInput so it wont happen?

src/errors/checkXcmTxInputs.ts
```diff
@@ -406,12 +462,14 @@ const checkSystemToSystemAssetId = (
  * @param xcmDirection
  * @param registry
  */
-export const checkAssetIdInput = (
+export const checkAssetIdInput = async (
```
TarikGul, line 465: Same with this function to make the api the first arg.

src/createXcmTypes/util/isValidForeignAssetMultiLocation.ts
```diff
@@ -0,0 +1,33 @@
...
+		}
+
+		return false;
+	} catch (error) {
+		const errorInfo = (error as Error).message.split('::');
+		const errorDetails = errorInfo[errorInfo.length - 2].concat(
+			errorInfo[errorInfo.length - 1]
+		);
+
+		throw new BaseError(`error creating MultiLocation type:${errorDetails}`);
```
TarikGul, line 31:
```suggestion
		throw new BaseError(`Error creating MultiLocation type:${errorDetails}`);
```

src/errors/checkLocalTxInputs.ts
```diff
@@ -17,52 +22,100 @@ enum LocalTxType {
...
 	specName: string,
-	registry: Registry
-): LocalTxType => {
-	const relayChainInfo = registry.currentRelayRegistry;
-	const systemChainId = getChainIdBySpecName(registry, specName);
-	const systemParachainInfo = relayChainInfo[systemChainId];
-
+	registry: Registry,
+	_api: ApiPromise,
+	transferForeignAssets: boolean | undefined
```
TarikGul, line 31: `transferForeignAssets: boolean` as well as the rename

src/createXcmCalls/limitedTeleportAssets.ts
IkerAlus: perhaps I am getting this wrong, but with this design, are we adding a new optional input to specify if the user wants to transfer a foreign asset? what is the motivation behind this decision? A user transferring a foreign asset should have the same UX as transferring native assets. The API can know it is a foreign asset as soon as the asset ID input is a multilocation and not a simple integer. Is it possible to avoid the extra optional value and have the same UX as for native assets? 

Note that this is different than with pool assets where the same ID system as normal assets is used and hence it may get confusing. Plus it is a way less critical functionality so it is ok to add the extra UX step.

src/createXcmTypes/util/getSystemChainAssetId.ts
marshacb: makes sense! I've updated all similar cases to have `api` as the first arg

src/createXcmTypes/util/getSystemChainAssetId.ts
marshacb: yes, the `isForeignAssetsTransfer` boolean is only true if the assetId is a `MultiLocation` as a string.

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
marshacb: Great question! This is in a try catch since if the `multilocation` string is invalid it will throw an error. This same function is used in both `checkLocalTxInput` and `checkXcmTxInput` to check if a `multilocation` asset exists and will error in either of them if the string is invalid. It's also being used in `getSystemChainAssetId` for local txs.

src/createXcmCalls/limitedTeleportAssets.ts
marshacb: yes, the UX will be the same for Assets and ForeignAssets for the user

src/createXcmTypes/util/getSystemChainAssetId.ts
TarikGul: Ahh gotcha! nice!

src/createXcmTypes/util/foreignAssetMultiLocationIsInRegistry.ts
TarikGul: Awesome :), Thanks for the explanation.

src/createXcmTypes/util/getSystemChainAssetId.ts
```diff
@@ -0,0 +1,99 @@
...
+ * errors if given an invalid symbol, integer id or multilocation that is
+ * not registered or found in the chains state
+ *
+ * @param tokenSymbol string
+ * @param specName string
+ */
+export const getSystemChainAssetId = async (
+	asset: string,
+	specName: string,
+	_api: ApiPromise,
```
TarikGul, line 23: This one still needs the API to be first.

marshacb: great catch! Updating now :)

src/AssetsTransferApi.spec.ts
```diff
@@ -157,13 +167,181 @@ describe('AssetTransferAPI', () => {
...
+					Direction.SystemToRelay,
+					AssetType.Native,
+					false,
+					systemAssetsApi.registry
+				);
+
+				expect(assetCallType).toEqual('Teleport');
+			});
+		});
+		describe('SystemToSystem', () => {
```
TarikGul, line 217: Web have 2 `describe`s here with `SystemToSystem` I think we can just have one with both `its` under it.

src/AssetsTransferApi.ts
```diff
@@ -408,7 +458,7 @@ export class AssetsTransferApi {
 		/**
 		 * Check if the origin is a parachain, and the destination is a system parachain.
 		 */
-		if (_api.query.polkadotXcm && isDestIdSystemPara) {
+		if (_api.query.polkadotXcm || (_api.query.xTokens && isDestIdSystemPara)) {
```
TarikGul, line 461: I think this needs to be `(_api.query.polkadotXcm || _api.query.xTokens) && isDestIdSystemPara)`.

src/AssetsTransferApi.ts
```diff
@@ -496,6 +549,127 @@ export class AssetsTransferApi {
...
+		) {
+			return AssetCallType.Teleport;
+		}
+
+		// para to para -> reserve
+		if (xcmDirection === Direction.ParaToPara) {
+			return AssetCallType.Reserve;
+		}
+
+		// para to system -> reserve !relay asset
```
TarikGul, line 648: What the `!relay` mean here.

src/createXcmTypes/util/assetIdsContainsRelayAsset.spec.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright 2023 Parity Technologies (UK) Ltd.
+
+import { Registry } from '../../registry';
+import { assetIdsContainRelayAsset } from './assetIdsContainsRelayAsset';
+
+describe('assetIdsContainsRelayAsset', () => {
+	// it('Should return true when assetIds contains relay chain symbol', () => {
```
TarikGul, line 7: This test is commented out, just as a note.

src/createXcmTypes/util/assetIdsContainsRelayAsset.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+		const assetIds = [
+			`{"parents": 1, "interior": {"Here": ''}}`,
+			`{"parents": 1, interior: {"X1": {"Parachain": "2023"}}}`,
+		];
+
+		const result = assetIdsContainRelayAsset(assetIds, registry);
+
+		expect(result).toEqual(true);
+	});
+	// it('Should return false when assetIds does not contain the relay assets multilocation', () => {
```
TarikGul, line 34: This test as well.

src/errors/checkLocalTxInputs.ts
```diff
@@ -17,52 +22,100 @@ enum LocalTxType {
...
-	}
-
-	let assetId = assetIds[0];
+	if (isForeignAssetsTransfer) {
+		if (assetIds.length === 0) {
+			throw new BaseError(
+				'Local foreignAsset transactions must have the `assetIds` input be a length of 1'
+			);
+		}
+		// check the foreignAssetsInfo to see if the given foreignAssetMatches
```
TarikGul, line 46: The comment here is a bit confusing specifically with `foreignAssetMatches`

src/createXcmTypes/util/multiLocationAssetIsParachainsNativeAsset.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright 2023 Parity Technologies (UK) Ltd.
+
+/**
+ * Given a multilocation assetId and destChainId, returns if the multilocation is the native asset of the destchain
```
TarikGul, line 4: The comments grammar can be fixed here.

src/util/getFeeAssetItemIndex.ts
```diff
@@ -1,8 +1,10 @@
 // Copyright 2023 Parity Technologies (UK) Ltd.
 
-import { getSystemChainTokenSymbolGeneralIndex } from '../createXcmTypes/util/getTokenSymbolGeneralIndex';
+import { ApiPromise } from '@polkadot/api';
+
+import { getChainAssetId } from '../createXcmTypes/util/getChainAssetId';
 import { MultiAsset } from '../types';
-import { NonRelayNativeInterior, RelayNativeInterior } from '../types';
+// import { NonRelayNativeInterior, RelayNativeInterior } from '../types';
```
TarikGul, line 7: Commented imports here.

TarikGul (approved):
Really solid job! Giving a thumbs up as I think the majority of this is ready. Just a few nits to clean up.

src/createXcmTypes/util/assetIdsContainsRelayAsset.spec.ts
marshacb: great catch. uncommenting

src/errors/checkLocalTxInputs.ts
marshacb: ill update this to be more clear

src/AssetsTransferApi.ts
marshacb: updating this comment

bitcoinfinancier (approved):

examples/systemToSystemXcmForeignAssets.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
IkerAlus, line 1: Collectives parachain does not have asset or foreignAsset pallets. This example only makes sense if sending DOT to Collectives.

examples/systemToSystemLocalForeignAssets.ts
```diff
@@ -0,0 +1,56 @@
...
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetsTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send foreign asset '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}'
+ * from a Kusama Asset Hub (System Parachain) account
+ * to a Kusama Asset Hub (System Chain) account, where the `xcmVersion` is set to 2, and the `isLimited` declaring that
```
IkerAlus, line 13: Should use in both cases "System Parachain" so the user does not think it is a  different chain

examples/systemToParaTeleportForeignAssets.ts
```diff
@@ -0,0 +1,56 @@
...
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetsTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send foreign asset '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}'
```
IkerAlus, line 11: Here I would clearly mention it will be a teleport because we are transferring a token native to the destination chain.

examples/systemToParaReserveTransferForeignAssets.ts
```diff
@@ -0,0 +1,56 @@
...
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetsTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send foreign asset '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}'
```
IkerAlus, line 11: This will be a teleport so the name of the file is misleading.

examples/paraToSystemForeignAssetsReserveTransferAssets.ts
```diff
@@ -0,0 +1,56 @@
...
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetsTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send foreign asset '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}'
```
IkerAlus, line 11: do we really want to create this example now? There is no foreign asset '{"parents":"1","interior":{"X2":[{"Parachain":"2125"},{"GeneralIndex":"0"}]}}' in Moonriver. It will be misleading for someone checking the example as of today.

IkerAlus (changes requested):
I have some comments regarding  the examples.

I will do the code review afterwards.

src/AssetsTransferApi.ts
```diff
@@ -496,6 +549,127 @@ export class AssetsTransferApi {
...
+
+		// system to system native asset -> teleport
+		if (
+			assetType === AssetType.Native &&
+			xcmDirection === Direction.SystemToSystem
+		) {
+			return AssetCallType.Teleport;
+		}
+
+		// system to system foreign asset -> reserve
```
IkerAlus, line 609: This case can't happen as of now in the ecosystem.

src/AssetsTransferApi.spec.ts
```diff
@@ -157,13 +167,179 @@ describe('AssetTransferAPI', () => {
...
+					['ksm'],
+					Direction.SystemToSystem,
+					AssetType.Native,
+					false,
+					systemAssetsApi.registry
+				);
+
+				expect(assetCallType).toEqual('Teleport');
+			});
+			it('Should correctly return Reserve when sending a foreign asset', () => {
```
IkerAlus, line 231: I am not totally sure about this assumption here. Also, at the moment there such scenario (sending foreign assets between system paras) is not possible in the ecosystem. I would remove this spec

src/createXcmTypes/util/sortMultiAssetsAscending.spec.ts
```diff
@@ -8,208 +9,1327 @@ describe('sortMultiAssetsAscending', () => {
...
+				id: {
+					Concrete: mockSystemApi.registry.createType('MultiLocation', {
+						parents: 1,
+						interior: mockSystemApi.registry.createType(
+							'InteriorMultiLocation',
+							{
+								X3: [
+									{ Parachain: '2000' },
+									{ PalletInstance: '50' },
+									{ GeneralIndex: '1984' },
```
IkerAlus, line 621: what happens in this scenario when asset 1 and asset 3 are the same? will the API just construct a call with two asset repeated? I think this is not permitted by the XCM format, the asset amounts should be added and placed in only one array element.

src/createXcmTypes/util/sortMultiAssetsAscending.spec.ts
```diff
@@ -8,208 +9,1327 @@ describe('sortMultiAssetsAscending', () => {
...
+
+		const res = sortMultiAssetsAscending(multiAssets);
+
+		expect(res.length).toEqual(expected.length);
+		expect(res[0].id).toEqual(expected[0].id);
+		expect(res[res.length - 1].fun).toEqual(expected[expected.length - 1].fun);
+		expect(JSON.stringify(res)).toEqual(JSON.stringify(expected));
+	});
+
+	it('Should correctly differnt X2 MultiLocations', () => {
```
IkerAlus, line 638: `it('Should correctly sort different X2 MultiLocations'`?

src/createXcmTypes/util/sortMultiAssetsAscending.spec.ts
```diff
@@ -8,208 +9,1327 @@ describe('sortMultiAssetsAscending', () => {
...
+
+		const res = sortMultiAssetsAscending(multiAssets);
+
+		expect(res.length).toEqual(expected.length);
+		expect(res[0].id).toEqual(expected[0].id);
+		expect(res[res.length - 1].fun).toEqual(expected[expected.length - 1].fun);
+		expect(JSON.stringify(res)).toEqual(JSON.stringify(expected));
+	});
+
+	it('Should correctly sort an unsorted multiasset array based on fungible value when all other values are the same', () => {
```
IkerAlus, line 1249: if all the other values are the same, is it not the same asset?

IkerAlus: the API should error out if the user input the same asset twice

examples/systemToSystemXcmForeignAssets.ts
marshacb: great catch! removing this example since we have one of sending the relay asset already.

examples/systemToParaReserveTransferForeignAssets.ts
marshacb: In this example we are sending `TNKR` from assethub to Moonriver which I assumed should be a reserve transfer since it isn't native the destination. Is that incorrect?

src/createXcmTypes/util/sortMultiAssetsAscending.spec.ts
marshacb: great catch! right now the api is checking for dupes at the `MultiAsset` level and not the assetId/MultiLocation level. Will update this to throw an error when duplicate `assetId`/ `MultiLocations` are passed in.

src/createXcmTypes/util/sortMultiAssetsAscending.spec.ts
marshacb: yep, same as above. Adding test cases for this covering assetId and MultiLocation dupes

src/errors/checkXcmTxInputs.ts
```diff
@@ -380,22 +552,145 @@ const checkParaToSystemAssetId = (
...
+			} else if (assetId.toLowerCase().includes('parents')) {
+				multiLocationAssetIdFound = assetId;
+			} else {
+				symbolAssetIdFound = assetId;
+			}
+		}
+
+		if (relayDefaultValueFound && multiLocationAssetIdFound) {
+			throw new BaseError(
+				`Found both default relay native asset and multilocation assetId: ${multiLocationAssetIdFound}. Asset Ids must be either be a combination of empty string, symbol and integer or multilocations exclusively.`
```
IkerAlus, line 661: Suggestion for error message: "Found both default relay native asset and foreign asset with assetId: ${multiLocationAssetIdFound}. Relay native asset and foreign assets can't be transferred within the same call."

src/errors/checkXcmTxInputs.ts
```diff
@@ -380,22 +552,145 @@ const checkParaToSystemAssetId = (
...
+			);
+		}
+
+		if (integerAssetIdFound && multiLocationAssetIdFound) {
+			throw new BaseError(
+				`Found both integer ${integerAssetIdFound} and multilocation assetId ${multiLocationAssetIdFound}. Asset Ids must be symbol and integer or multilocation exclusively.`
```
IkerAlus, line 667: Suggestion for error message: "Found both native asset with assetID ${integerAssetIdFound} and foreign asset with assetId ${multiLocationAssetIdFound}. Native assets and foreign assets can't be transferred within the same call."

src/errors/checkXcmTxInputs.ts
```diff
@@ -380,22 +552,145 @@ const checkParaToSystemAssetId = (
...
+			);
+		}
+
+		if (symbolAssetIdFound && multiLocationAssetIdFound) {
+			throw new BaseError(
+				`Found both symbol ${symbolAssetIdFound} and multilocation assetId ${multiLocationAssetIdFound}. Asset Ids must be symbol and integer or multilocation exclusively.`
```
IkerAlus, line 673: is this true in general? what if the registry has also foreign assets by symbol (which I assume should be the case in the future)?

src/errors/checkXcmTxInputs.ts
IkerAlus: same comment for the spec file

src/errors/checkXcmTxInputs.ts
IkerAlus: same comment for the spec file

src/createXcmTypes/util/sortMultiAssetsAscending.ts
```diff
@@ -13,32 +12,190 @@ import {
  */
 export const sortMultiAssetsAscending = (multiAssets: MultiAsset[]) => {
```
IkerAlus, line 13: It is likely that future versions of XCM will add new junction types, different arrays etc and this function is not going to be easy to maintain. We should have this into consideration if we want to keep this implementation of the sorting function.

IkerAlus (changes requested):
Just small suggestions, and an open discussion point about the sorting function. Great and solid work.

src/errors/checkXcmTxInputs.ts
marshacb: the current way we determine if an asset is a foreign asset is if it is passed in as a multilocation. If we expect that symbols will eventually be present in the foreign assets pallet as well then we would need to change our implementation to account for this scenario

IkerAlus: that is a good point. I expect foreign assets to be identified by symbols in the same way as any other asset. But lets leave this functionality for later. I think it is ok for now to leave it like this

IkerAlus (approved):
LGTM! Great work! 

P.S: what is the status regarding the last to-do point "add foreign assets pallet instance from registry"? does that one belong to the registry repo?

marshacb:
> LGTM! Great work!
> 
> P.S: what is the status regarding the last to-do point "add foreign assets pallet instance from registry"? does that one belong to the registry repo?

Yes, I've updated the registry to fetch the foreignAssetsPalletInstance already. After I do a release of the registry this morning it will be a simple change of pulling the pallet instance from there rather than using a constant.

Issue #221, opened 2023-07-12
Support for `xTokens` for `ParaToSystem`


TarikGul:
Not all parachains support calls for `polkadotXcm` even if the pallet does exist. Therefore we need to add support for `xTokens` so we can differentiate between the two. 

## Proposed Idea

- Inside of `createXcmCalls` we need to have 2 folders: `polkadotXcm`, and `xTokens`. We should take the existing calls and put them under `polkadotXcm` as that is where they belong. 

- Implement support for:
    - [transfer_multiasset](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L241)
    - [transfer_multiasset_with_fee](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L313)
    - [transfer_multiassets](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/xtokens/src/lib.rs#L375)

- Note that since this is only for `ParaToSystem` we only need to add the methods to `ICreateXcmType`, we will need to do some typescript magic there to ensure that is the only direction that allows us to have these calls.

PR #227, opened 2023-07-25
feat: add support for xTokens pallet

marshacb:
TODO:

- [x] unit tests
- [x] integration tests
- [x] error validation layer
- [x] update examples
- [x] docs
- [x] fetch xcAsset registry/update asset-transfer-api-registry with xcAssets

closes: [#221](https://github.com/paritytech/asset-transfer-api/issues/221)


src/createXcmCalls/polkadotXcm/limitedTeleportAssets.ts
```diff
@@ -31,7 +31,9 @@ export const limitedTeleportAssets = (
 	xcmVersion: number,
 	specName: string,
 	registry: Registry,
-	weightLimit?: string,
+	isLimited?: boolean,
```
TarikGul, line 34: I think we should move away from adding a bunch of optional params at the end of the args, and instead put them in a single option.

src/createXcmTypes/RelayToSystem.ts
```diff
@@ -148,14 +148,24 @@ export const RelayToSystem: ICreateXcmType = {
 	 * Create a WeightLimitV2 type.
```
TarikGul, line 148: If we use `XcmV3WeightLimit`, lets make sure to change the header comment to reflect that.

TarikGul: Same with all the other `createWeightLimit` changes.

src/AssetsTransferApi.spec.ts
```diff
@@ -659,6 +659,8 @@ describe('AssetTransferAPI', () => {
 					xcmVersion: 3,
 					weightLimit: '100000',
 					isLimited: true,
+					refTime: '1000',
```
TarikGul, line 662: Note this should be updated with the Weightlimit change.

src/createXcmTypes/util/getAssetHubAssetId.ts
```diff
@@ -0,0 +1,128 @@
...
+import { ApiPromise } from '@polkadot/api';
+
+import { ASSET_HUB_CHAIN_ID } from '../../consts';
+import { BaseError } from '../../errors';
+import { Registry } from '../../registry';
+import { foreignAssetMultiLocationIsInRegistry } from './foreignAssetMultiLocationIsInRegistry';
+import { getChainIdBySpecName } from './getChainIdBySpecName';
+
+/**
+ * Returns the correct asset id for a valid AssetHub token symbol integer id
```
TarikGul, line 12: This comment is a bit confusing. I think the punctuation just needs to be fixed and it will be perfect.

src/AssetsTransferApi.ts
```diff
@@ -278,8 +294,13 @@ export class AssetsTransferApi {
 			assetIds,
 			amounts,
 			xcmDirection,
+			xcmPallet,
 			_specName,
 			registry,
+			{
```
TarikGul, line 300: If this is an option should it be at the end of the args? And I think`xcmVersion`, and `paysWithFeeDest` can be added to it.

TarikGul (approved):
Rockstart PR! Really impressive!

src/AssetsTransferApi.ts
marshacb: good catch. adding those to the options

PR #240, opened 2023-08-11
feat: add poolAssets support

TarikGul:
## Summary

The following PR adds support for liquid token transfers. 

### Option - `transferLiquidToken`

This option allows for a user to send one or more liquid tokens within the assetIds field. This works for both crosschain transfers and local transfers.

```typescript
/**
 * Boolean to declare if this will transfer liquidity tokens.
 * Default is false.
 */
transferLiquidToken?: boolean;
```

### Errors and Validation

- All tokens must be liquid tokens when the `transferLiquidToken` option is true.
    It will error with the following error if it is incorrect:
    ```
    No liquid token asset was detected. When setting the option "transferLiquidToken" to true a valid liquid token assetId must be present.
    ```
- The asset ids must all be integers are strings.
    It will error with the following error if it is incorrect:
    ```
    Liquid Tokens must be valid Integers
    ```
    
### Tests

I added tests in the following places:

- Integration tests
- Check Xcm Tx Input validation layer
- Added mock test information to add a single asset under `mockApi.query.poolAssets.asset()` which is asset number `0`
- Changed system metadata from asset-hub-kusama to asset-hub-westend


src/testHelpers/adjustedMockSystemApi.ts
```diff
@@ -110,46 +137,17 @@ const asset = (assetId: number): Promise<Option<PalletAssetsAssetDetails>> =>
...
+			multiLocationAssetInfo
 		);
 		assets.set(1984, sufficientAsset);
 
 		const maybeAsset = assets.has(assetId) ? assets.get(assetId) : undefined;
 
 		if (maybeAsset) {
 			return new Option(
-				createStatemineRegistry(9430),
+				createStatemineRegistry(9435),
```
marshacb, line 150: I know using westend asset hub is temporary. Should we change this name to reflect that in the meantime?

marshacb (approved):
great work on this PR. LGTM! Left a small question and gave it a check :)

examples/localLpToken.ts
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * import { AssetsTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetsTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send LiquidPool Asset '0' from a Westmint (System Parachain) account
+ * to a Injected (Parachain) called 'testing', where the `xcmVersion` is set to 2, and the `isLimited` declaring that
```
marshacb, line 12: We should update the description to mention this is a local transfer

TarikGul: resolved

src/testHelpers/adjustedMockSystemApi.ts
TarikGul: Hmm this is a good question. Im down to keep it as statemine, since it technically doesn't matter. 

But also I am kinda torn on this.

IkerAlus (approved):
LGTM, great work!

Issue #218, opened 2023-07-06
Change `SYSTEM_PARACHAIN_IDS`


TarikGul:
Currently we hardcode the `SYSTEM_PARACHAIN_IDS`, but we can safely assume that any `para_id` that is less than 2000 and greater than 0 will be a system parachain. 

https://github.com/paritytech/asset-transfer-api/blob/main/src/consts.ts#L19

PR #247, opened 2023-08-16
Update System Parachain Id check based on chainId

marshacb:
* removed SYSTEM_PARACHAINS_IDS
* checks if a chainId is a system chain based on whether its ID is strictly less than 2000 and greater than 0

closes: [#218](https://github.com/paritytech/asset-transfer-api/issues/218)

### DIFF
--- a/src/errors/checkXcmTxInputs.ts
+++ b/src/errors/checkXcmTxInputs.ts
@@ -5,9 +5,9 @@ import { isHex } from '@polkadot/util';
 import { isEthereumAddress } from '@polkadot/util-crypto';
 
 import {
+	ASSET_HUB_CHAIN_ID,
 	MAX_ASSETS_FOR_TRANSFER,
 	RELAY_CHAIN_IDS,
-	SYSTEM_PARACHAINS_IDS,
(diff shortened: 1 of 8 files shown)


src/errors/checkXcmTxInputs.ts
```diff
@@ -544,16 +544,16 @@ export const checkIsValidSystemChainAssetId = async (
...
 	api: ApiPromise,
 	assetId: string,
 	relayChainInfo: ChainInfo,
 	registry: Registry,
 	isForeignAssetsTransfer: boolean
 ) => {
-	const systemParachainId = SYSTEM_PARACHAINS_IDS[0];
-	const systemParachainInfo = relayChainInfo[systemParachainId];
-	const systemSpecName = systemParachainInfo.specName;
+	const assetHubChainId = ASSET_HUB_CHAIN_ID;
```
TarikGul, line 554: Little nit, but I think we can remove this and just do `relayChainInfo[ASSET_HUB_CHAIN_ID];`

TarikGul (approved):
LGTM, just one little nit

PR #252, opened 2023-08-22
fix(docs): update README & docs

Imod7:
#### Small fixes
- Updated the examples link found [here](https://github.com/paritytech/asset-transfer-api/tree/main#docs) so it works both from the [README](https://github.com/paritytech/asset-transfer-api/tree/main#docs) and the [docs](https://paritytech.github.io/asset-transfer-api/).
- Replaced the checkmarks & cross in the [Cross-Chain Support](https://github.com/paritytech/asset-transfer-api/tree/main#current-cross-chain-support) section so it appears correctly both in the README and the docs

#### Version Update in the docs
- Updated the release version to `v0.1.0-beta.3` in the docs
   - I just changed the version from `.2` to `.3` in one of the instances in the `index.html` file, e.g. [here](https://github.com/paritytech/asset-transfer-api/blob/9ed7ddd0500685d347780bdd8df48c7fe8ec49d7/docs/index.html#L14)
   - run `npx typedoc`
   - and it updated all the files in the docs where the version is mentioned.


TarikGul (approved):
LGTM, thanks so much!

marshacb (approved):

PR #256, opened 2023-08-23
[WIP]: add cache to registry

marshacb:


TarikGul:
From an initial look. I think the registry should have implemented getters and setters that either look for a given item in the cache, or set an item in the cache. 

For example:

```
class Registry {
    constructor() {
        ....
        this.cache = {
            ...
        }
    }
    
    public cacheLookupAsset(assetId: string) {
        
    }
    
    public cacheAddAsset(assetId: string) {
    
    }
}
```

NOTE: We already know the specname, and the relaychain so we dont need those values as args.

marshacb:
> From an initial look. I think the registry should have implemented getters and setters that either look for a given item in the cache, or set an item in the cache.
> 
> For example:
> 
> ```
> class Registry {
>     constructor() {
>         ....
>         this.cache = {
>             ...
>         }
>     }
>     
>     public cacheLookupAsset(assetId: string) {
>         
>     }
>     
>     public cacheAddAsset(assetId: string) {
>     
>     }
> }
> ```
> 
> NOTE: We already know the specname, and the relaychain so we dont need those values as args.

great idea! Ill modify this to reflect the getter setter structure

TarikGul:
Some additional thoughts. 

1. For setting items in the `cache` since we have specific types it would either need to take in a second `arg` called `type` (or something like it) that tells us whether the assetId is a LP token, FA token, or just regular asset part of the assets pallet. 

OR

2. We can have multiple methods,  `setForeignAssetInCache`, `setLiquidPoolTokenInCache`, etc. That way each `assetId` can be a different format>

OR 

3. We can be more generic for (2), and have a single function that has two args: 

```typescript
type GenericAsset<T> = T extends 'foreignAsset'
	? <some_type>
	: T extends 'liquidPools'
	? <some_type>
	: T extends 'assetpallet'
	? <some_type>
	: never;
	
...
public setAssetIdInCache<T>(asset: GenericAsset<T>, type: T)
...
```

src/createXcmTypes/util/getAssetId.ts
```diff
@@ -0,0 +1,168 @@
...
+				assetId = asset;
+				const assetMetadata = await _api.query.assets.metadata(asset);
+				const assetSymbol = assetMetadata.symbol.toHuman()?.toString();
+
+				if (assetSymbol) {
+					// add queried asset to registry
+					registry.setAssetInCache(asset, assetSymbol);
+				}
+			} else {
+				throw new BaseError(`general index for assetId ${asset} was not found`);
```
TarikGul, line 120: BaseError now takes in a second arg to specify what kind of error it is. You can view them at `/errors/BaseError.ts`

src/createXcmTypes/util/getAssetId.ts
```diff
@@ -0,0 +1,168 @@
...
+			}
+		} else {
+			throw new BaseError(
```
TarikGul, line 123: Same as above.

src/createXcmTypes/util/getAssetId.ts
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	// check the registry and return the assetId if found
+	// currently the registry does not contain assetsInfo for parachains
+	// so we skip checking them
+	const { assetsInfo } = registry.currentRelayRegistry[currentChainId];
+	if (!isParachain) {
+		if (Object.keys(assetsInfo).length === 0) {
+			throw new BaseError(
```
TarikGul, line 52: Same comment as the one below for BaseError second arg

src/createXcmTypes/util/getAssetId.ts
```diff
@@ -0,0 +1,168 @@
...
+ * @param asset string
+ * @param specName string
+ * @param isForeignAssetsTransfer boolean
+ */
+export const getAssetId = async (
+	_api: ApiPromise,
+	registry: Registry,
+	asset: string,
+	specName: string,
+	isForeignAssetsTransfer?: boolean
```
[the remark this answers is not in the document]
TarikGul: Im assuming its not applicable here?

src/registry/Registry.ts
```diff
@@ -14,13 +16,140 @@ export class Registry {
...
+			kusama: {},
+			westend: {},
+			xcAssets: {
+				polkadot: [],
+				kusama: [],
+			},
+		};
+	}
+
+	private initializeCurrentChainIdCache() {
```
TarikGul, line 38: This can be called directly in the constructor so you dont have to call it anywhere else, and it will only be called once.

src/createXcmTypes/util/foreignAssetMultiLocationIsInCacheOrRegistry.ts
```diff
@@ -2,30 +2,54 @@
...
 
-export const foreignAssetMultiLocationIsInRegistry = (
+export const foreignAssetMultiLocationIsInCacheOrRegistry = (
 	api: ApiPromise,
 	multilocationStr: string,
 	registry: Registry
 ): boolean => {
-	try {
-		const assetHubChainId = 1000;
+	if (!registry.cache[registry.relayChain][ASSET_HUB_CHAIN_ID]) {
```
TarikGul, line 15: This should be added to the registry layer as part of the initialization call in the constructor.

src/registry/Registry.ts
```diff
@@ -14,13 +16,140 @@ export class Registry {
...
+			return this.cache[this.relayChain][currentChainId]['assetsInfo'][assetId];
+		}
+
+		return undefined;
+	}
+
+	/**
+	 * Setter for the assets cache.
+	 *
+	 * @param assetId string
```
TarikGul, line 145: This is a bit confusing between `assetId`, and `asset`. A better description differentiating between the both would be nice.

src/registry/Registry.ts
```diff
@@ -14,13 +16,140 @@ export class Registry {
...
+			] as ForeignAssetsData;
+		}
+
+		return undefined;
+	}
+
+	/**
+	 * Setter for the foreignAssetsInfo cache.
+	 *
+	 * @param assetId string
```
TarikGul, line 78: Same as below for the description.

src/errors/checkXcmTxInputs.ts
```diff
@@ -436,64 +474,78 @@ const checkSystemAssets = async (
...
 		// check if assetId is a number
 		const parsedAssetIdAsNumber = Number.parseInt(assetId);
 		const invalidNumber = Number.isNaN(parsedAssetIdAsNumber);
 
 		if (!invalidNumber) {
-			// assetId is a valid number
-			// ensure the assetId exists as an asset on the system parachain
-			const assetSymbol: string | undefined =
-				systemParachainInfo.assetsInfo[assetId];
+			let assetSymbol: string | undefined = undefined;
```
TarikGul, line 484:
```suggestion
			let assetSymbol: string | undefined;
```

src/errors/checkXcmTxInputs.ts
```diff
@@ -436,64 +474,78 @@ const checkSystemAssets = async (
...
+
+			// check the cache for the asset
+			// if not in cache, check the registry
+			if (registry.cacheLookupAsset(assetId)) {
+				assetSymbol = registry.cacheLookupAsset(assetId);
+			} else if (systemParachainInfo.assetsInfo[assetId]) {
+				assetSymbol = systemParachainInfo.assetsInfo[assetId];
+			}
 
 			if (assetSymbol === undefined) {
```
TarikGul, line 494:
```suggestion
			if (!assetSymbol) {
```

src/errors/checkXcmTxInputs.ts
```diff
@@ -436,64 +474,78 @@ const checkSystemAssets = async (
...
+					const assetStr = assetSymbol as string;
+					// add the asset to the cache
+					registry.setAssetInCache(assetId, assetStr);
 				}
 			}
 		} else {
 			// not a valid number
 			// check if id is a valid token symbol of the system parachain chain
-			let isValidTokenSymbol = false;
+			if (assetId === '') {
```
TarikGul, line 519: Is this suppose to error instead of return?

TarikGul (changes requested):
Just some nits to address on the above. But overall really solid PR!

src/errors/checkXcmTxInputs.ts
```diff
@@ -436,64 +474,78 @@ const checkSystemAssets = async (
...
 				if (asset.isNone) {
 					throw new BaseError(
 						`(${xcmDirection}) integer assetId ${assetId} not found in ${specName}`,
 						BaseErrorsEnum.AssetNotFound
 					);
+				} else {
+					const assetSymbol = (
+						await api.query.assets.metadata(parsedAssetIdAsNumber)
+					).symbol
+						.toHuman()
```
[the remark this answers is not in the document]
marshacb: toString() returns the hex string of the metadata field. So toHuman() returns human readable form as AnyJson then we toString() it to get the string representation of that.

src/errors/checkXcmTxInputs.ts
marshacb: since we allow `''` to represent the relay token,  here I've made it return early since it's valid. It will throw an error if all of the assetId checks fail/the asset isn't found or if multiple non unique matches of the given asset are found. Is there a different pattern you think works better here?

src/registry/Registry.ts
marshacb: very good point. Ill make it so we call this in the constructor instead

src/errors/checkXcmTxInputs.ts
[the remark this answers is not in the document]
TarikGul: Ahh okay makes sense, thanks for the explanation!

src/errors/checkXcmTxInputs.ts
TarikGul: Ahh perfect makes sense. Thanks for the explanation!

src/createXcmTypes/util/getAssetId.ts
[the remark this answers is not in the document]
marshacb: yes, I don't think we've passed `isLiquidPoolAssetTransfer` as of now. I think because they are always expected to be integers corresponding to the lpToken integer asset ID

src/createXcmTypes/util/foreignAssetsMultiLocationExists.ts
```diff
@@ -3,9 +3,12 @@
 import { ApiPromise } from '@polkadot/api';
 
 import { BaseError, BaseErrorsEnum } from '../../errors';
-
+import { Registry } from '../../registry';
+import { ForeignAssetsData } from '../../registry/types';
+import { AssetMetadata } from '../../types';
```
TarikGul, line 8:
```suggestion
import type { AssetMetadata } from '../../types';
```

src/createXcmTypes/util/getAssetId.ts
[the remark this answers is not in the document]
TarikGul: Makes perfect sense thanks :)

TarikGul (approved):
Some small nits, but really amazing job!

PR #260, opened 2023-08-24
feat: `sendersAddr`, add options validation, and fix payload format

TarikGul:
## Summary

- Fix payload format for addresses by adding a new option called `sendersAddr`
- Add a validation layer for options

### `sendersAddr`

This is required when the passed in `format` option for `createTransferTransaction` is set to `payload`.


marshacb (approved):
Great PR. LGTM!

Issue #263, opened 2023-08-29
Remove `Number.parseInt` and replace with BN.js


TarikGul:
Currently we use `Number.parseInt` to check for valid numbers but this may fail if an Asset is above javascripts MAX_VALUE. Therefore we should replace checking a numbers validity as a string with BN.js

PR #264, opened 2023-08-30
fix: replace all unsafe instances of parseInt

TarikGul:
closes: https://github.com/paritytech/asset-transfer-api/issues/263

This replaces all instances where a MAX_INTEGER_OVERFLOW can happen.

### DIFF
--- a/src/errors/checkXcmTxInputs.ts
+++ b/src/errors/checkXcmTxInputs.ts
@@ -15,6 +15,7 @@ import { Registry } from '../registry';
 import type { ChainInfo, ChainInfoKeys } from '../registry/types';
 import { XCMChainInfoKeys } from '../registry/types';
 import { AssetInfo, Direction } from '../types';
+import { validateNumber } from '../validate';
 import { BaseError, BaseErrorsEnum } from './BaseError';
 
 /**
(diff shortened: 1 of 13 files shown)


src/errors/checkXcmTxInputs.ts
```diff
@@ -666,11 +658,12 @@ export const checkParaAssets = async (
 	} else {
 		// not a valid number
 		// check if id is a valid token symbol of the system parachain chain
-		if (!invalidNumber) {
+		// TODO: Im confused about this, clarify this logic.
```
TarikGul, line 661: @marshacb Lets go over this in a call because I am not sure I understand the conditional below.

Will that if statement ever be hit? I think we might have talked about this in the past but it would be nice to have a refresher.

TarikGul: If you look at the diff to the left, if it doesnt enter that first `if` it will never hit the second one inside of the `else` statement. Or atleast that is what I understand here.

src/validate/validateNumber.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+		const res = validateNumber('123456789000000');
+
+		expect(res).toEqual(true);
+	});
+	it('Should return false on an invalid number as a string with a length less than 16', () => {
+		const res = validateNumber('test');
+
+		expect(res).toEqual(false);
+	});
+	it('Should correctly validate an number as a string with a length greater than 16', () => {
```
marshacb, line 14: same as above

marshacb (approved):
LGTM! just a very minor nit

src/errors/checkXcmTxInputs.ts
TarikGul: We figured out what we need to do in a conversation and a PR will be up shortly with the fixes

PR #265, opened 2023-08-30
fix: unused conditional, and DRY cache logic in checkParaAssets

TarikGul:


marshacb (approved):
awesome! looks good :)

Issue #268, opened 2023-08-31
Access `foreignAssetsPalletInstance` from the metadata instead of the registry


TarikGul:
Similar to how we do it with: [fetchPalletInstanceId](https://github.com/paritytech/asset-transfer-api/blob/main/src/createXcmTypes/util/fetchPalletInstanceId.ts)

We should do the same with any instance of `foreignAssetsPalletInstance`. Therefore we can remove that logic from the registry, and slim it down.

TarikGul:
Once we update the registry, and remove the last remaining pieces of data from the `registry.spec.ts` files then we can close this issue.

TarikGul:
Completed via https://github.com/paritytech/asset-transfer-api/pull/284

PR #270, opened 2023-08-31
fix: adjust `fetchPalletInstanceId` to handle `ForeignAssets` pallet

TarikGul:
rel: https://github.com/paritytech/asset-transfer-api/issues/268

This PR focuses on giving support for `ForeignAssets` inside of `fetchPalletInstanceId` so we can remove `foreignAssetsPalletInstance` from the registry.


marshacb (approved):
LGTM!

PR #275, opened 2023-09-12
feat: add disabled options config and function

TarikGul:
## Internal functionality

This ensures and allows us to add disabling for certain options regarding the `createTransferTransaction` function ie `TransferArgsOpts<T extends Format>`. All one needs to do is going into `config/disabledOpts` and switch the `disabled` key to true, and add which chains are disabled in the `chains` field. 

NOTE: for the `chains` field if we need to disable it for all chains just add `*`.

### DIFF
--- a/src/errors/disableOpts.ts
+++ b/src/errors/disableOpts.ts
@@ -1,5 +1,6 @@
 // Copyright 2023 Parity Technologies (UK) Ltd.
 
+import type { MappedOpts } from '../config/disabledOpts';
 import { disabledOpts } from '../config/disabledOpts';
 import type { Format, TransferArgsOpts } from '../types';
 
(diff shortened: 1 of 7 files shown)


src/errors/disableOpts.ts
```diff
@@ -0,0 +1,31 @@
...
+export const disableOpts = <T extends Format>(
+	opts: TransferArgsOpts<T>,
+	specName: string
+) => {
+	const optKeys = Object.keys(opts) as MappedOpts[];
+	const chain = specName.toLowerCase();
+
+	for (const key of optKeys) {
+		const disabledKeyInfo = disabledOpts[key];
+		if (opts[key] && disabledKeyInfo.disabled) {
```
marshacb, line 22: will `opts[key]` ever be undefined here?

TarikGul: Yes totally possible, which then means there is no option to disable. 

Example: Lets say the user doesn't input `paysWithFeeOrigin`, then it wont even check if its disabled as it wont matter.

marshacb (approved):
Awesome PR! Just have one small question. LGTM!

src/errors/disableOpts.ts
marshacb: makes total sense!

PR #276, opened 2023-09-13
fix: remove remaining use of `parseInt`

TarikGul:
This removes all the remaining instances of `parseInt`. It also chains the input of `isSystemChain` to only take in strings for `chainId`.


marshacb (approved):
Really nice changes! LGTM!

PR #284, opened 2023-09-15
fix: update the registry to the new xcAssets format

TarikGul:
## Summary

This PR is an extension of https://github.com/paritytech/asset-transfer-api-registry/pull/57. It focuses on reducing the size of the registry, while also removing the `xcAssets` key from the registry and correctly adding that data to each parachains key within the registry. This PR is also an extension of https://github.com/paritytech/asset-transfer-api/pull/292 where I remove the `MultiLocation` type creation in favor of using `resolveMultiLocation` where it specifies the exact versioned multilocation needed for each type creation. 

### Registry

**Previous**
```json
{
    polkadot: {}
    kusama: {}
    westend: {}
    xcAssets: {}
}
```

**Current**
```
{
    polkadot: {
        '2000': {
            ...
            xcAssetsData[]
        }
    }
    kusama: {}
    westend: {}
}
```

### End notes

This PR will be followed up with a PR that adds Rococo to the registry.


Imod7 (approved):
LGTM! Quite impressive refactoring! 💯

bee344 (approved):
Good job!

PR #288, opened 2023-09-19
fix: remove `getChainIdBySpecName` and add caching system

TarikGul:
This removes the util function `getChainIdBySpecName` and replaces it with a class method in the registry called `lookupChainIdBySpecName` which now caches values that have been searched.

Part 2 of this PR will handle the empty string which is returned by adding a validation layer for the destId, and originId


bee344 (approved):
Great job

PR #293, opened 2023-09-22
feat: add rococo support and to the registry

TarikGul:
This adds rococo support to the registry

closes: https://github.com/paritytech/asset-transfer-api/issues/290

PR #297, opened 2023-09-26
fix: rococo initialization in the registry

TarikGul:
rel: https://github.com/paritytech/asset-transfer-api/pull/293

This ensures that the registry can initialize with rococo as the relay chain.

## Important note:

In order to use rococo's asset hub with the `AssetTransferApi` one will need to hardcode the `specName` into the initialization like the following:

```typescript
new AssetTransferApi(api, `asset-hub-rococo`, xcmVersion);
```

The reason being, kusama's asset hub and and rococo's asset-hub both share the same specName currently and will cause conflicts. We currently do an overewrite in the registry that the api uses and set the `specName` for rococo's asset hub to be `asset-hub-rococo`. This should be resolved at a higher level in the below PR for asset-hub's spec.

rel: https://github.com/paritytech/polkadot-sdk/pull/1215

### DIFF
--- a/examples/rococoAssetHubToRelay.ts
+++ b/examples/rococoAssetHubToRelay.ts
@@ -1,0 +1,50 @@
+/**
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetTransferApi, constructApiPromise } from '../src';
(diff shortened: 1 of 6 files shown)


TarikGul:
Im running into one issue that I can see will be a potential bug. 

The specName for `rococo`s asset-hub is also statemine... Which would cause a conflict in lookup.

Imod7 (approved):

PR #300, opened 2023-10-11
feat: support para to para xcm tx construction

marshacb:


README.md
```diff
@@ -28,19 +28,23 @@ Find the full documentation [here](https://paritytech.github.io/asset-transfer-a
...
 | Direction              | V2                 | V3                 |
 | ---------------------  | ------------------ | ------------------ |
 | System to Parachain    |         ✅         |      ✅            |
 | System to Relay        |         ✅         |      ✅            |
 | Relay to Parachain     |         ✅         |      ✅            |
 | Relay to System        |         ✅         |      ✅            |
-| Parachain to Parachain |         ❌         |      ❌            |
-| Parachain to Relay     |         ❌         |      ❌            |
+| Parachain to Parachain |         ✅         |      ❌            |
+| Parachain to Relay     |         ❌         |      ❎            |
```
TarikGul, line 40: I think we need to put it on the para to para line, but i also think we can just do the Check mark for now, and write a note under explaining the situations we dont cover.

examples/paraToParaParachainPrimaryNativeReserveTransfer.ts
```diff
@@ -0,0 +1,44 @@
...
+ * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send PHA from a Phala Network (Parachain) account
+ * to a Moonbeam (Parachain) account, where the `xcmVersion` is set to 3, and `isLimited` is false declaring that
+ * it will allow `unlimited` weight for the tx.
```
TarikGul, line 13: This is kinda true, if isLimited is `true` it will be unlimited unless `weightLimit` is provided

TarikGul: If isLimited is false it moreso means that it will use `reserveTransferAssets` or `teleport` instead of `limited*`

examples/paraToParaTransferMultiAsset.ts
```diff
@@ -0,0 +1,44 @@
...
+import { AssetTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send BNC from a Moonriver (Parachain) account
+ * to a Bifrost Kusama (Parachain) account, where the `xcmVersion` is set to 2, and `isLimited` is false declaring that
+ * it will allow `unlimited` weight for the tx.
+ *
+ * NOTE: When `isLimited` is true it will expect for refTime and proofSize to be provided as additional arguments.
```
TarikGul, line 15: Same with this as above, if `isLimited` is true and no `weightLimit` is passed in, it will be unlimited.

examples/paraToParaTransferMultiAssets.ts
```diff
@@ -0,0 +1,44 @@
...
+ * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send vMOVR and vBNC from a Moonriver (Parachain) account
+ * to a Bifrost Kusama (Parachain) account, where the `xcmVersion` is set to 2, and the `isLimited` is false declaring that
+ * it will allow `unlimited` weight for the tx.
```
TarikGul, line 13: Same as above.

src/AssetTransferApi.ts
```diff
@@ -539,22 +555,32 @@ export class AssetTransferApi {
 	 * @param destChainId
 	 * @param specName
 	 */
-	private establishDirection(destChainId: string, specName: string, destIsSystemParachain: boolean): Direction {
```
TarikGul, line 542: Nice refactor!

src/AssetTransferApi.ts
```diff
@@ -294,8 +307,11 @@ export class AssetTransferApi {
 		let txMethod: Methods;
 		let transaction: SubmittableExtrinsic<'promise', ISubmittableResult>;
 
-		if (xcmPallet === XcmPalletName.xTokens && xcmDirection === Direction.ParaToSystem) {
-			if (paysWithFeeDest && !paysWithFeeDest.includes('parents') && assetIds.length < 2) {
+		if (
+			xcmPallet === XcmPalletName.xTokens &&
+			(xcmDirection === Direction.ParaToSystem || xcmDirection === Direction.ParaToPara)
+		) {
+			if (!paysWithFeeDest && assetIds.length < 2) {
```
TarikGul, line 314: Can we do assetIds.length < 2 in one of our layer checks? Maybe in `checkBaseInputTypes`?

src/AssetTransferApi.ts
```diff
@@ -364,7 +380,7 @@ export class AssetTransferApi {
 				txMethod = 'limitedReserveTransferAssets';
 				transaction = await limitedReserveTransferAssets(
 					_api,
-					xcmDirection,
+					xcmDirection as XcmDirection,
```
TarikGul, line 383: I think we should avoid having to type cast here. I would assume its the `Direction.local` which is causing this issue, we might want to find a way to change that. More comments below.

src/AssetTransferApi.ts
```diff
@@ -539,22 +555,32 @@ export class AssetTransferApi {
 	 * @param destChainId
 	 * @param specName
 	 */
-	private establishDirection(destChainId: string, specName: string, destIsSystemParachain: boolean): Direction {
+	private establishDirection(
+		isLocal: boolean,
```
TarikGul, line 559: My thought would be to remove the `isLocal` check and keep that seperate, since `establishDirection` is really focusing on xcmDirections, and would help us from not having to typeCast. 

And when we need to check for the localTx we just do `if (isLocalTx) ...`

TarikGul: That being said I am not deadset on removing `isLocal` happy to hear some counter thoughts.

src/createXcmCalls/util/establishXcmPallet.ts
```diff
@@ -20,21 +20,14 @@ export enum XcmPalletName {
  */
 export const establishXcmPallet = (
 	api: ApiPromise,
-	direction?: Direction,
+	direction?: Direction | string,
```
TarikGul, line 23: Why is `string` possible here?

README.md
marshacb: Good eye. fixed!

examples/paraToParaParachainPrimaryNativeReserveTransfer.ts
marshacb: good point! Ive updated these to match the way we describe it in other examples to make it more consistent

src/createXcmCalls/util/establishXcmPallet.ts
marshacb: shouldn't be possible. fixed!

src/AssetTransferApi.ts
marshacb: Great question! After looking at it a bit I don't think it would work as an alternative place to check the assetIds length since in this case we aren't checking the length for validity but more as a way to determine which `xTokens` method to use. Definitely open to discuss it more though

TarikGul: Ahh great point, that totally makes sense. Lets keep the check as it is then! Thanks for the explanation.

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+			const beneficiary = ParaToPara.createBeneficiary(
+				mockParachainApi,
+				'0xf5d5714c084c112843aca74f8c498da06cc5a2d63153b825189baa51043b1f0b',
+				2
+			);
+
+			const expectedRes = {
+				v2: {
+					parents: 0,
+					interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 20: Do we need InteriorMultiLocation here? Can it just be the object. When I did the refactor a while ago I remember that InteriorMultiLocation also resolves only to V2 types, so it could pose issues in the future, especially for V3.

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+			const beneficiary = ParaToPara.createBeneficiary(
+				mockParachainApi,
+				'0xf5d5714c084c112843aca74f8c498da06cc5a2d63153b825189baa51043b1f0b',
+				3
+			);
+
+			const expectedRes = {
+				v3: {
+					parents: 0,
+					interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 45: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+		});
+	});
+	describe('Destination', () => {
+		it('Should work for V2', () => {
+			const destination = ParaToPara.createDest(mockParachainApi, '100', 2);
+
+			const expectedRes = {
+				v2: {
+					parents: 1,
+					interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 66: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+
+			expect(destination.toJSON()?.toString()).toStrictEqual(expectedRes.toString());
+		});
+		it('Should work for V3', () => {
+			const destination = ParaToPara.createDest(mockParachainApi, '100', 3);
+
+			const expectedRes = {
+				v3: {
+					parents: 1,
+					interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 82: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+				}
+			);
+
+			const expectedRes = {
+				v2: [
+					{
+						id: {
+							concrete: {
+								parents: 1,
+								interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 116: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+						},
+						fun: {
+							fungible: 1000000000000,
+						},
+					},
+					{
+						id: {
+							concrete: {
+								parents: 1,
+								interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 129: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+				}
+			);
+
+			const expectedRes = {
+				v3: [
+					{
+						id: {
+							concrete: {
+								parents: 1,
+								interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 163: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+						},
+						fun: {
+							fungible: 1000000,
+						},
+					},
+					{
+						id: {
+							concrete: {
+								parents: 1,
+								interior: mockParachainApi.registry.createType('InteriorMultiLocation', {
```
TarikGul, line 176: Same here for `InteriorMultiLocation`

src/createXcmTypes/ParaToPara.ts
```diff
@@ -0,0 +1,499 @@
...
+		const { registry, paysWithFeeDest, specName, assetIds, amounts, xcmVersion } = opts;
+		if (xcmVersion && xcmVersion === 3 && specName && amounts && assetIds && paysWithFeeDest) {
+			const multiAssets = await createParaToParaMultiAssets(
+				api,
+				normalizeArrToStr(amounts),
+				specName,
+				assetIds,
+				xcmVersion,
+				registry,
+				opts.isForeignAssetsTransfer
```
TarikGul, line 200: lets also destructure `isForeignAssetsTransfer`.

src/createXcmTypes/ParaToPara.ts
```diff
@@ -0,0 +1,499 @@
...
+			);
+
+			const assetIndex = getFeeAssetItemIndex(
+				api,
+				registry,
+				paysWithFeeDest,
+				multiAssets,
+				specName,
+				xcmVersion,
+				opts.isForeignAssetsTransfer
```
TarikGul, line 210: Same here as above.

TarikGul (commented):
Incredible PR. Just some more comments, once resolved it's a thumbs up from me!

src/AssetTransferApi.ts
marshacb: The reason I decided to add a `Local` `Direction` is because currently for txs that are `local balance transfers` on the relay chain, `establishDirection` returns the direction `RelayToPara`. There is no `RelayToRelay` and in the future this description won't necessarily be correct to treat as a `Local` relay chain transaction since there will be multiple Relay chains communicating with each other. So we could either allow the default for this case to continue to be `RelayToPara` so that `establishDirection` doesn't throw an error being unable to determine a direction or we could go with a solution similar to this. Open to discuss it more either way

TarikGul: Ahh i see the reason, okay I think your implementation is sound than and totally works for me!

TarikGul (approved):
lgtm, just a few more nits to sort out in the comments but once those are good this is good to go in. Lets do a live test of this as well.

Issue #301, opened 2023-10-20
Refactor type generation


TarikGul:
rel: https://github.com/polkadot-js/api/issues/5741#issuecomment-1772151935

Given the above comment, we should try and remove as much type creation as we can and let the call generation do most of the heavy lifting.

- [x] Dest, and Beneficiary
- [x] Assets
- [x] WeightLimit
- [x] XTokens
- [x] Resolve multilocation, and others

TarikGul:
For `Assets` which I am working on now, here are some notes:

1. First and foremost we need a more centralized mapping of the XCM types that do not come from pjs. Which means we don't want the Codec types, and type creation that is necessary to use those types. We want a more primitive version that can map directly to whatever JSON or object we use. So this needs to be mapped out for V2, and V3.
- [ ] Is completed?

2. Remove `resolveMultiLocation`.

3. Update each XcmDirection with the updated types that will be returned from the MultiAsset creation, and update the calls to take in the correct types. This will also require removing the `createType` call in each instance.

4. Update `sortMultiAssetsAscending`. This is by far the most challenging one, but should be easy to verify because all of the tests are written already.

TarikGul:
This is now complete, the last type of refactor we can do is with the 'Calls'

TarikGul:
Completed via: https://github.com/paritytech/asset-transfer-api/pull/313

PR #303, opened 2023-10-20
fix(internal): refactor dest and beneficiary types generators

TarikGul:

part 1.

This ensures we don't rely on `createType` for type generation, and focuses first on refactoring destination and beneficiary type generators to ensure PR's stay small.


bee344 (approved):
Lgtm

PR #305, opened 2023-10-23
fix(internal): Refactor multiassets creation, and xcm types

TarikGul:
## Summary

This PR focuses on refactoring all xcm generation for assets to be local to the api, and not involve polkadot-js. This means there is no longer the use to use `registry.createType` for any xcm generation.

The main focuses are:

- Refactor and reorg the xcm types to be all local.
- add `sanitizeKeys`.
- Refactor `createAssets` for all directions. This may mean helper and utility functions were all changed as well.
- Refactor XTokens creation.
- Remove unnecessary tests and functions that did not have any meaning anymore.
- Refactor `sortMultiAssetsAscending` to only use local xcm types.


marshacb (approved):
Nice refactor! LGTM

PR #310, opened 2023-11-02
fix(internal): refactor all weight limits to not use createType

TarikGul:
## Summary

- Refactor all `createWeightLimit` calls to be return `XcmWeight` types instead of Polkadot-js weight types.
- Refactor all `createFeeAssetItem` calls to return a `number` type instead of a `u32` polkadot-js type.


marshacb (approved):
lgtm

PR #312, opened 2023-11-03
test: add liquid assets script to test network

bee344:
Two questions I have regarding this:
* Currently I have to force pjs to accept `<any>` when dealing with the `asset-conversion-pallet`, because the types it expects come from the `substrate-node` implementation and are incompatible with the `Kusama Asset Hub` implementation, but due to this I have to disable `eslint` for those two lines.
```typescript
const createLiquidityPoolCall = (api: ApiPromise) => {
	// For now, we have to override the types of the Assets until PJS is updated
	// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
	return api.tx.assetConversion.createPool(<any>native, <any>asset);
};

const addLiquidityCall = (api: ApiPromise, amountNative: number, amountAsset: number, to: KeyringPair) => {
	// For now, we have to override the types of the Assets until PJS is updated
	// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
	return api.tx.assetConversion.addLiquidity(<any>native, <any>asset, amountNative, amountAsset, 0, 0, to.address);
};
```
* Trappist doesn't support `asset-conversion` so all the process runs on KAH, wondering how we feel about that.

**Edit**: [Issue](https://github.com/paritytech/trappist/issues/281) to add asset-conversion to trappist.
**Edit 2**: [Types](https://github.com/polkadot-js/api/blob/7a8a4f1c64ce5dbb5bf56558054c7e906e81d133/packages/types-augment/src/lookup/types-substrate.ts#L5229) on PJS that are used on Asset Conversion

### DIFF
--- a/scripts/testNetworkLiquidAssets.ts
+++ b/scripts/testNetworkLiquidAssets.ts
@@ -2,135 +2,156 @@
 
 import { ApiPromise, WsProvider } from '@polkadot/api';
 import { Keyring } from '@polkadot/keyring';
-import { KeyringPair } from '@polkadot/keyring/types'
+import { KeyringPair } from '@polkadot/keyring/types';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
 import chalk from 'chalk';
 
 import { KUSAMA_ASSET_HUB_WS_URL } from './consts';
 import { awaitBlockProduction, delay, logWithDate } from './util';
 
-
 const ASSET_ID = 1;
-const ASSET_NAME = "Testy";
-const ASSET_TICKER = "TSTY";
+const ASSET_NAME = 'Testy';
+const ASSET_TICKER = 'TSTY';
 const ASSET_DECIMALS = 12;
 const ASSET_MIN = 1;
 
 const asset = {
-    parents: 0,
-    interior: {
-        X2: [
-            { palletInstance: 50 },
-            { generalIndex: ASSET_ID },
-        ]
-    }
-
+	parents: 0,
+	interior: {
+		X2: [{ palletInstance: 50 }, { generalIndex: ASSET_ID }],
+	},
 };
 
 const native = {
-    parents: 1,
-    interior: {
-        Here: '',
-    },
+	parents: 1,
+	interior: {
+		Here: '',
+	},
 };
 
 const createAssetCall = (api: ApiPromise, admin: KeyringPair) => {
-    return api.tx.assets.create(ASSET_ID, admin.address, ASSET_MIN)
+	return api.tx.assets.create(ASSET_ID, admin.address, ASSET_MIN);
 };
 
 const setMetadataCall = (api: ApiPromise) => {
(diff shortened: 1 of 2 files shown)


scripts/testNetworkLiquidAssets.ts
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const mintCall = (api: ApiPromise, to: KeyringPair, amount: number) => {
+	return api.tx.assets.mint(ASSET_ID, to.address, amount);
+};
+
+const createLiquidityPoolCall = (api: ApiPromise) => {
+	// For now, we have to override the types of the Assets until PJS is updated
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+	return api.tx.assetConversion.createPool(<any>native, <any>asset);
```
TarikGul, line 47: This is totally fine with me. I double checked the ts compilter issue happening with PJs and this is the right approach.

scripts/testNetworkLiquidAssets.ts
```diff
@@ -0,0 +1,157 @@
...
+};
+
+const addLiquidityCall = (api: ApiPromise, amountNative: number, amountAsset: number, to: KeyringPair) => {
+	// For now, we have to override the types of the Assets until PJS is updated
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+	return api.tx.assetConversion.addLiquidity(<any>native, <any>asset, amountNative, amountAsset, 0, 0, to.address);
```
TarikGul, line 53: Same with this, totally fine.

TarikGul:
> Trappist doesn't support asset-conversion so all the process runs on KAH, wondering how we feel about that.

I think this should be totally fine, not much we can do. That being said if we can still create the liquid pool assets thats a step forward. In this case though we just can't transfer them to Trappist if I understand this correctly?

TarikGul (approved):
Overall, amazing script. LGTM!

bee344:
> > Trappist doesn't support asset-conversion so all the process runs on KAH, wondering how we feel about that.
> 
> I think this should be totally fine, not much we can do. That being said if we can still create the liquid pool assets thats a step forward. In this case though we just can't transfer them to Trappist if I understand this correctly?

Correct, right now there's no way to send the LP tokens to Trappist, but we create them just fine and we can even send them between accounts, so that's ok.

TarikGul:
> > > Trappist doesn't support asset-conversion so all the process runs on KAH, wondering how we feel about that.
> > 
> > 
> > I think this should be totally fine, not much we can do. That being said if we can still create the liquid pool assets thats a step forward. In this case though we just can't transfer them to Trappist if I understand this correctly?
> 
> Correct, right now there's no way to send the LP tokens to Trappist, but we create them just fine and we can even send them between accounts, so that's ok.

Okay I think that's perfect then and the script does exactly what it intends on doing. Feel free to merge it in, whenever your'e ready

PR #313, opened 2023-11-06
fix(internal): remove createType(Call) where applicable

TarikGul:
Remove `createType('Call')` where applicable, and add a eslint disable in the scripts


marshacb (approved):
nice. lgtm!

PR #314, opened 2023-11-06
test: fix jest tests running twice

TarikGul:
Since the refactor of the createTypes, I had fixed the examples `build` for tsconfig, which caused the tests to run twice. This PR ensures that the `/build` folder is ignored for tests.


bee344 (approved):
LGTM

PR #315, opened 2023-11-06
fix: add Ethereum address check in createBeneficiary for ParaToPara direction 

marshacb:


src/createXcmTypes/ParaToSystem.ts
```diff
@@ -44,21 +45,27 @@ export const ParaToSystem: ICreateXcmType = {
 	 */
 	createBeneficiary: (accountId: string, xcmVersion?: number): XcmDestBenificiary => {
 		if (xcmVersion == 2) {
+			const X1 = isEthereumAddress(accountId)
```
TarikGul, line 48: Can a system parachain have a `ethereum` address?

marshacb: ahh, good point. Im guessing that's not possible currently. Will revert changing this direction

TarikGul: Sounds good! I wasn't sure if there is a possible type of system parachain that has an ethereum address, but probably best to be safe, and remove it.

PR #317, opened 2023-11-07
fix: xtokens and xTokens naming

TarikGul:
This ensures we check and supply the right naming of `xTokens` pallet depending on the chain we are dealing with.


marshacb (approved):
nice! lgtm

PR #321, opened 2023-11-10
feat: ParaToRelay

TarikGul:
## Summary

Make `ParaToRelay` available as a xcm direction for transaction construction. Most parachains use `xTokens` therefore it gives support for both `xTokens`, and `polkadotXcm`.

closes: https://github.com/paritytech/asset-transfer-api/issues/211


examples/paraToRelayTransferMultiAsset.ts
```diff
@@ -0,0 +1,44 @@
...
+ * When importing from @substrate/asset-transfer-api it would look like the following
+ *
+ * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
+ */
+import { AssetTransferApi, constructApiPromise } from '../src';
+import { TxResult } from '../src/types';
+import { GREEN, PURPLE, RESET } from './colors';
+
+/**
+ * In this example we are creating a call to send MOVR from a Moonriver (Parachain) account
```
marshacb, line 11:
```suggestion
 * In this example we are creating a call to send KSM from a Moonriver (Parachain) account
```

examples/paraToRelayTransferMultiAsset.ts
```diff
@@ -0,0 +1,44 @@
...
+ * to Kusama Asset Hub (Relay chain) account, where the `xcmVersion` is set to 3, and `isLimited` is false declaring that
```
marshacb, line 12:
```suggestion
 * to a Kusama Relay chain account, where the `xcmVersion` is set to 3, and `isLimited` is false declaring that
```

src/AssetTransferApi.spec.ts
```diff
@@ -336,6 +336,22 @@ describe('AssetTransferAPI', () => {
 					moonriverAssetsApi.registry
 				);
 
+				expect(assetCallType).toEqual('Reserve');
+			});
+		});
+		describe('ParaToRelay', () => {
+			it('Should correctly return XTokens', () => {
```
marshacb, line 343: should we have`return Reserve` here in the `it` or do we mean to add `xTokens` as a new fetchCallType?

src/AssetTransferApi.ts
```diff
@@ -306,12 +306,13 @@ export class AssetTransferApi {
...
-
 		if (
 			(xcmPallet === XcmPalletName.xTokens || xcmPallet === XcmPalletName.xtokens) &&
-			(xcmDirection === Direction.ParaToSystem || xcmDirection === Direction.ParaToPara)
+			(xcmDirection === Direction.ParaToSystem ||
+				xcmDirection === Direction.ParaToPara ||
+				xcmDirection === Direction.ParaToRelay)
 		) {
-			if (!paysWithFeeDest && assetIds.length < 2) {
+			if (xcmDirection === Direction.ParaToRelay || (!paysWithFeeDest && assetIds.length < 2)) {
```
marshacb, line 315: Since we check that `xcmDirection` is possibly `ParaToRelay` on line `313` is the additional `xcmDirection === Direction.ParaToRelay` equality check needed?

marshacb (approved):
awesome PR. Just a few minor nits and a question but this looks great to me!

src/AssetTransferApi.spec.ts
TarikGul: Ahh yes, you're right it should return `Reserve`, good catch

src/AssetTransferApi.ts
TarikGul: So the reason I did this was to enforce that the TX was always `transferMultiAsset`.

Just incase so if a fee portion was added it will still resolve to `transferMultiAsset`.

marshacb: nice! that makes sense

TarikGul: I'll write a comment there to clarify though!

PR #323, opened 2023-11-10
fix: support parachains without assets pallet in runtime

marshacb:
* Handle case when a parachain does not include the Assets pallet in its runtime to allow tx construction instead of throwing an error
* Refactor `ParaTo*` direction integration tests and separate into different files
* Add support for Ethereum AccountKey20 addresses when creating `xTokens` beneficiaries for `ParaToPara` direction
* Rename `transferMultiAsset*` to `transferMultiasset*`
* Replace references to onfinality rpcs with blastapi in `ParaTo*` examples

closes: [#320](https://github.com/paritytech/asset-transfer-api/issues/320)


package.json
```diff
@@ -13,7 +13,7 @@
     "start:zombienet-liquidity-assets-script": "yarn build:scripts && node ./scripts/build/testNetworkLiquidAssets.js",
     "lint": "substrate-dev-run-lint",
     "lint:fix": "substrate-dev-run-lint --fix",
-    "test": "NODE_ENV=test substrate-exec-jest --detectOpenHandles",
+    "test": "NODE_ENV=test substrate-exec-jest --silent --detectOpenHandles",
```
TarikGul, line 16: Was this pushed up accidently?

TarikGul (commented):
I'll have a closer look at this on Monday. But it looks super solid.

package.json
marshacb: good catch. Was using during debugging/testing

src/createXcmTypes/util/fetchPalletInstanceId.ts
```diff
@@ -11,6 +11,10 @@ import { BaseError, BaseErrorsEnum } from '../../errors';
  * @param isLiquidToken Boolean to determine whether or not to fetch the PoolAssets id.
  */
 export const fetchPalletInstanceId = (api: ApiPromise, isLiquidToken: boolean, isForeignAsset: boolean): string => {
+	if (!api.query.assets) {
```
TarikGul, line 14: DQ: But if the assets pallet is not available than we can assume that `PoolAssets`, and `ForeignAssets` will be unavailable as well?

TarikGul (approved):
LGTM, really solid!

I really like the integrationTests abstraction. Makes things way more organized and easier to work through! Just one small question, but good to merge once the conflict is resolved

src/createXcmTypes/util/fetchPalletInstanceId.ts
marshacb: thats a great point. As is we would not check whether it is a pool assets or foreign assets tx if the assets pallet is not supported. I've updated this in order to evaluate the `poolAssets` and `foreignAssets` pallet cases first. We will only return an empty string prior to filtering the pallet names if the assets pallet isn't available and the palletName is an empty string.

TarikGul: Sounds perfect! Thanks for the response. Looks like a solid change!

PR #327, opened 2023-11-17
chore: use cdn hosted assets registry

marshacb:
* removes the `@substrate/asset-transfer-api-registry` npm package
* constructs the ATA registry by fetching the latest registry data from `https://paritytech.github.io/asset-transfer-api-registry/registry.json`

### DIFF
--- a/examples/paraToParaParachainPrimaryNativeReserveTransfer.ts
+++ b/examples/paraToParaParachainPrimaryNativeReserveTransfer.ts
@@ -4,6 +4,7 @@
  * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
  */
 import { AssetTransferApi, constructApiPromise } from '../src';
+import { parseRegistry, Registry } from '../src/registry';
 import { TxResult } from '../src/types';
 import { GREEN, PURPLE, RESET } from './colors';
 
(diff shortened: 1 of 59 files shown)


TarikGul:
One concern I have is I see the registry keeps updating every hour and is deleting and readding a few parachains.

I am looking into this, but this shouldn't go in until that is resolved.

TarikGul:
We also need a way for users to opt out of the cdn registry and opt in for the static npm registry

TarikGul:
I would also change the commit name, since this is a fundamental change to the API this is going to break things. It's more of a `fix!`.

On the topic of fundamental breaking changes. We can avoid breaking the API all together by initializing the registry once the first call is made. Basically the first time `createTransferTransaction` is called we see if the registry has been initialized, if it hasn't we set it, and if it has we do nothing.

That way the API stays super simple for the user, and we don't apply any breaking changes.

TarikGul (changes requested):
Overall, I think we should avoid adding the registry as a constructor input since it adds complexity for the user, and we should maintain handling everything internally. (Suggestion above)

There should be an opt in, and opt out option that allows the user to choose where the registry comes from as well, whether its fetched from the CDN or pulled in from the npm package.

Addition: I know in private conversation I agreed to the current implementation, but after some review I think we should be careful adding any breaking changes to the core public facing functionality and DX unless it is absolutely necessary or we introducing a whole new API in general.

PR #332, opened 2023-11-28
fix: give registry option between CDN and npm package

TarikGul:
replaces: https://github.com/paritytech/asset-transfer-api/pull/327


marshacb (approved):
lgtm

PR #336, opened 2023-12-15
fix: check fee asset lp exists

marshacb:
* adds a check to ensure that a provided `paysWithFeeOrigin` multilocation has an existing liquidity pool


TarikGul (approved):
LGTM, really nice job. 

2 nits:

1. Docs will have to be updated since the `BaseErrorEnum` now has the new error.

2. I think the conventional commit can be changed to `fix: ...` instead of `test: ...`

Issue #357, opened 2024-01-29
Injected registry should allow for overriding and deep comparison


TarikGul:
Currently the `injectedRegsitry` option just allows for adding in a chain based on ID. But it actually doesn't do any deep comparison.

Example:

If the registry contains

```
polkadot: {
    '9876': {
        tokens: ['TST'],
	assetsInfo: { ... },
	foreignAssetsInfo: { ... },
	specName: 'testing',
	poolPairsInfo: { ... },
    },
},
```

The user should be able to input:

```
injectedRegistry: {
    polkadot: {
        '9876': {
            tokens: ['TST', 'TST2']
        }
    }
}
```

Which will only update the tokens field with the new values. This should be applicable to any field in the registry.

bee344:
This is partially covered by #359 , which covers the update and deep comparison. The overriding is still pending

TarikGul:
Completed via: https://github.com/paritytech/asset-transfer-api/pull/369

PR #359, opened 2024-02-01
feat: ensure the injectedRegsitry opt does deep comparisons

bee344:
### Description
It allows updating the entries of the existing registry with new info~, as well as overriding existing entries~. The injected registry no longer requires to have all the fields filled when updating an existing entry, but it errors out when adding a new entry and not specifying the chain's `specName` ~or `tokens`~.

#### Added `updateRegistry` to `parseRegistry`
Added the function `updateRegistry` to handle checking whether the entry already exists or if it's new info, and if it's already present, it updates the corresponding fields.

#### Added alternative interfaces with optional values
Added ~`InjectedChainInfoRegistry` , `InjectedChainInfo` and~ `InjectedChainInfoKeys` alternative interface~s~ to allow for injected registries with partial info, as well as  ~the `AssetTransferApiInjectedOpts` type for comparison~ made `AssetTransferApiOpts`, `ChainInfo` and `ChainInfoRegistry` to take either `InjectedChainInfoKeys` or `ChainInfoKeys` depending on what's needed.

#### Updated tests and `adjustedMockSystemApi`
Foreign assets' tests failed because `TNKR` is already in the registry, so I modified the `adjustedMockSystemApi`'s `foreignAssets` to another asset, updated the corresponding tests, and updated the `parseRegistry`'s tests to account for the new functionality.

Rel: #357


src/registry/parseRegistry.spec.ts
```diff
@@ -46,4 +46,124 @@ describe('parseRegistry', () => {
...
+		const registry = parseRegistry(reg as ChainInfoRegistry, opts);
+
+		expect(registry.rococo['2000']).toStrictEqual({
+			tokens: ['TST'],
+			assetsInfo: {},
+			foreignAssetsInfo: {},
+			specName: 'testy',
+			poolPairsInfo: {},
+		});
+	});
```
TarikGul, line 168: What if the current chain has 50 entries for `foreignAssetsInfo`, and we just want to add one more?

TarikGul: Is the current case we would have to copy all 50 entries in memory to another object, then add our new asset then override the registries version?

src/registry/parseRegistry.ts
```diff
@@ -1,21 +1,44 @@
...
-export const parseRegistry = (registry: ChainInfoRegistry, assetsOpts: AssetTransferApiOpts): ChainInfoRegistry => {
+const updateRegistry = (injectedChain: InjectedChainInfo, registry: ChainInfoRegistry, registryChain: string) => {
+	for (const key of Object.keys(injectedChain)) {
+		const info = registry[registryChain] as unknown as ChainInfo;
+		if (info[key] !== undefined) {
+			Object.assign(info[key], injectedChain[key]);
+		} else {
+			for (const property of Object.keys(info[0])) {
+				if (injectedChain[key][property] === undefined) {
+					if (property === 'specName' || property === 'tokens') {
```
TarikGul, line 15: Why are we singling out specName, and tokens here, what about the rest of the keys?

src/types.ts
```diff
@@ -161,6 +161,18 @@ export type AssetTransferApiOpts = {
 	registryType?: RegistryTypes;
 };
 
+export type AssetTransferApiInjectedOpts = {
```
TarikGul, line 164: We dont need to make a copy of the `AssetTransferApiOpts` type just use Generics to pass in a type that's necessary.

src/registry/Registry.ts
```diff
@@ -271,7 +271,7 @@ export class Registry {
 		for (let i = 0; i < paraIds.length; i++) {
 			const id = paraIds[i];
 			const chain = this.currentRelayRegistry[id];
-			if (chain.specName.toLowerCase() === specName.toLowerCase()) {
+			if (chain.specName !== undefined && chain.specName.toLowerCase() === specName.toLowerCase()) {
```
TarikGul, line 274:
```suggestion
			if (chain.specName && chain.specName.toLowerCase() === specName.toLowerCase()) {
```

TarikGul (changes requested):
On the first review I have some nits, and questions. So far though, amazing job!

src/registry/parseRegistry.ts
bee344: Because in this case we are adding a chain to the registry, and it must have `specName` and `tokens`, the other fields are autofilled with `{}` as per line 18.

bee344: I added a new test where you can see this at work

TarikGul:
As mentioned in an offline discussion we are moving this PR from a override to injection only, and a less rigid set of rules, making it as much as a deep comparison as possible

src/registry/parseRegistry.spec.ts
```diff
@@ -46,4 +46,586 @@ describe('parseRegistry', () => {
...
+						tokens: tokens,
+						assetsInfo: assetsInfo,
+					},
+				},
+			},
+		};
+
+		const registry = parseRegistry(reg as ChainInfoRegistry, opts);
+
+		expect(registry.kusama['1000'].assetsInfo).toStrictEqual({
```
TarikGul, line 194: This kind of check is unmaintainable, since the registry assets will grow, and we will have to update this test everytime we update the package. Instead just see if the value exists. Same with the tests below.

src/registry/parseRegistry.ts
```diff
@@ -1,21 +1,59 @@
...
+// Copyright 2023-2024 Parity Technologies (UK) Ltd.
 
 import { ASSET_HUB_CHAIN_ID } from '../consts';
-import type { AssetTransferApiOpts } from '../types';
-import type { ChainInfoRegistry } from './types';
+import type { AssetTransferApiInjectedOpts } from '../types';
+import type { ChainInfo, ChainInfoRegistry, InjectedChainInfo } from './types';
 
-export const parseRegistry = (registry: ChainInfoRegistry, assetsOpts: AssetTransferApiOpts): ChainInfoRegistry => {
+const updateRegistry = (injectedChain: InjectedChainInfo, registry: ChainInfoRegistry, registryChain: string) => {
```
TarikGul, line 7: For a cleaner solution we should really use recursion here for deep comparisons and updates. Having a triple nested loop is not ideal.

TarikGul: That being said there are arguments to both sides:

recursion: We could implement it once in a generic way, and never have to touch it again, even if the structure of the registry changes.

iteration: It's generally faster, and less memory intensive especially in stack frames. That being said we would have to update it every time there is a change in the registry.

So I am definitely up for discussion on this.

marshacb (approved):
lgtm!

src/registry/parseRegistry.ts
```diff
@@ -1,21 +1,108 @@
...
+	const buffer: ChainInfoKeys = {
+		tokens: [],
+		assetsInfo: {},
+		foreignAssetsInfo: {},
+		poolPairsInfo: {},
+		specName: '',
+	};
+	for (const id of Object.keys(injectedChain)) {
+		if (!chain[id] && !injectedChain[id].specName) {
+			throw Error(BaseErrorsEnum.SpecNameNotProvided);
```
TarikGul, line 81: For this lets use the BaseError function we created in the library, so:

```
throw new BaseError('<error_message>', BaseErrorsEnum.SpecNameNotProvided);
```

TarikGul (changes requested):
Just one little nit, then this is good to go! Really great job :)

src/registry/parseRegistry.ts
```diff
@@ -78,7 +78,7 @@ const updateRegistry = (
 	};
 	for (const id of Object.keys(injectedChain)) {
 		if (!chain[id] && !injectedChain[id].specName) {
-			throw Error(BaseErrorsEnum.SpecNameNotProvided);
+			throw new BaseError('<error_message>', BaseErrorsEnum.SpecNameNotProvided);
```
TarikGul, line 81: Replace the `<error_message>` with a actual error message

TarikGul (approved):
LGTM, really amazing job!

PR #361, opened 2024-02-05
questions [DO NOT MERGE]

TarikGul:


src/registry/parseRegistry.ts
```diff
@@ -4,17 +4,20 @@ import { ASSET_HUB_CHAIN_ID } from '../consts';
 import type { AssetTransferApiOpts } from '../types';
 import type { ChainInfo, ChainInfoKeys, ChainInfoRegistry, InjectedChainInfoKeys } from './types';
 
-const propertyIterator = (input: object, chain: ChainInfo<ChainInfoKeys>, id: string, property: string) => {
+// Question what if the input is an array filled with object that we need to deep check.
+const propertyIterator = (input: object, chain: ChainInfo<ChainInfoKeys>, id: string, property?: string) => {
```
TarikGul, line 8: With input I think we can get a better type than just an `object`, what if we want to pass in an array that is filled with objects?

PR #369, opened 2024-02-09
feat: allow for registry override

bee344:
### Description
Updated `parseRegistry` to allow for overriding entries of the existing registry. This is achieved by adding the optional property `overridingRegistry` to `AssetTransferApiOpts`. Now `propertyIterator` takes an extra optional argument `override`, a `boolean` that defines whether we intend to override the registry or not. This value is set on `parseRegistry` and depends on the presence of the property `overridingRegistry` in the `opts` passed to the function. If the `chainId` passed in `overridingRegistry` is not present in the registry, it just adds it as if it were an `injectedRegistry`. If both options are present, it first updates the registry with `injectedRegistry` and then makes the changes as per `overridingRegistry`.


src/registry/parseRegistry.spec.ts
```diff
@@ -469,73 +483,287 @@ describe('parseRegistry', () => {
...
+						tokens: ['WOP', 'WER'],
+					},
+				},
+			},
+		};
+		const registry = parseRegistry(reg as ChainInfoRegistry<ChainInfoKeys>, opts);
+
+		expect(registry.westend['0'].tokens).toStrictEqual(['WOP', 'WER']);
+	});
+	it("Should correctly override the registry's token entry", () => {
```
marshacb, line 570: duplicate test case

marshacb (approved):
minor nit. LGTM

src/types.ts
```diff
@@ -154,6 +154,10 @@ export type AssetTransferApiOpts<T extends ChainInfoKeys | InjectedChainInfoKeys
 	 * Option to inject chain information into the registry.
 	 */
 	injectedRegistry?: RequireAtLeastOne<ChainInfoRegistry<T>>;
+	/**
+	 * Option to override the registry with the supplied chain information.
+	 */
+	overridingRegistry?: RequireAtLeastOne<ChainInfoRegistry<InjectedChainInfoKeys>>;
```
TarikGul, line 160: To keep things in the active tense, I think we should call this `overrideRegistry`.

TarikGul (changes requested):
Overall looks amazing! Small nit, I think we should change the name to `overrideRegistry`. In addition to that we need to update the docs in the README.md that go over the `AssetTransferApiOpts`.

TarikGul (approved):
LGTM, great job :)

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/parity-tech-update-for-february/6630/1

Issue #370, opened 2024-02-13
Support XCM v4 new types


IkerAlus:
The new version of the XCM format, v4, has been released in Polkadot SDK v1.7 (https://github.com/paritytech/polkadot-sdk/pull/1230). The main change comes as renaming of all the basic types. 

Asset Transfer API should be ready for these changes as soon as this is live on a testnet.  Bear in mind that, we may have a situation where production networks are using the old types (XCM v3) for some time while testnets are already in XCM v4.

PR #372, opened 2024-02-21
fix: add support for xcmv4 types

marshacb:
### Description

This PR adds support for relevant type changes that were introduced in [XCMv4](https://github.com/paritytech/polkadot-sdk/pull/1230). 

### Changes
* Added v4 as a supported XCM version
* Added XCMv4 Asset type which no longer specifies `Concrete` as part of the AssetId
* Updated Destination/Beneficiary type construction to reflect new array junction syntax
* Added Westend -> AssetHub, AssetHub -> Westend and Asset Hub -> Collectives Xcmv4 examples
* Updated relevant type construction in `createXcmTypes`
* Added V4 test cases for each direction in `createXcmTypes`
* Removed `Multi` from `sortMultiAssetsAscending` and `dedupeMultiAssets`

closes: [#370](https://github.com/paritytech/asset-transfer-api/issues/370)


IkerAlus:
can we add a bit context here to help reviewers, also link to the issue this is supposed to close?

src/consts.ts
```diff
@@ -46,7 +46,7 @@ export const DEFAULT_XCM_VERSION = 2;
  * There should only ever be two supported versions.
  * Therefore supported xcm versions should have a fixed length of 2
```
TarikGul, line 47: Let's fix up the comment here, saying the length. will now always be of 3 since we have 3 versions. Once v5 comes out we can remove v2 etc.

TarikGul (approved):
Amazing Job LGTM, just 1 small nit

Polkadot-Forum:
This pull request has been mentioned on **Polkadot Forum**. There might be relevant details there:

https://forum.polkadot.network/t/parity-tech-update-for-february/6630/1

Issue #376, opened 2024-03-01
fix: remove pallet instance from foreign asset construction


marshacb:
remove the unneeded addition of pallet instance junctions from foreign asset locations

PR #377, opened 2024-03-01
fix: update foreign asset location construction

marshacb:
### Description

This PR fixes a bug where foreign asset locations had their `PalletInstance` prepended to them.
This is incorrect behavior which should be removed.

### Changes
* removes `constructForeignAssetMultilocation` function from the API
* adds additional foreign asset examples for local, reserve and teleport txs
* fixes test in `resolveMultiLocation` to require the correct xcm version for the `GlobalConsensus` type

closes: [#376](https://github.com/paritytech/asset-transfer-api/issues/376)


examples/kusama/assetHub/foreignAssetTransfers/local/assetHubForeignAssetsTransfer.ts
```diff
@@ -28,8 +27,6 @@ const main = async () => {
 			['1000000000000'],
 			{
 				format: 'call',
```
bee344, line 29:
```suggestion
				format: 'call',
				xcmVersion: 3
```
This example fails otherwise

marshacb: Great catch! After considering allowing this to be optional I think it makes sense to be explicit so updated the example rather than change it

bee344 (approved):
Great work!

Issue #386, opened 2024-03-30
Always use `limited`


joepetrowski:
The non-`limited` teleport/reserve transfer calls are deprecated. Instead, you should use `limited_*` with the weight limit set to `Unlimited`.

Ref: https://github.com/paritytech/asset-transfer-api/blob/main/examples/assetHubToMoonriverPaysWithFeeOrigin.ts#L17C2-L17C127

PR #392, opened 2024-04-02
fix!: always use limited calls

marshacb:
### Description

This PR removes support for creating `reserveTransferAssets` and `teleportAssets` calls which are now deprecated. Calls will now default to use `limitedReserveTransferAssets` or `limitedTeleportAssets` calls when appropriate.

This PR also removes the `isLimited` option when creating transfer transactions. In order to input a custom weight users now only need to provide a `weightLimit` option with their preferred `refTime` and `proofSize` values.

closes: [#386](https://github.com/paritytech/asset-transfer-api/issues/386)


TarikGul (changes requested):
- Should we get rid of `isLimited` all together?

- The README also needs to be updated.

marshacb:
> * Should we get rid of `isLimited` all together?
> * The README also needs to be updated.

We could remove `isLimited` and just consider whether the `weightLimit` option is provided to specify custom weights. Might be a simpler experience that way so Ill update with this in mind.

TarikGul:
> > * Should we get rid of `isLimited` all together?
> > * The README also needs to be updated.
> 
> We could remove `isLimited` and just consider whether the `weightLimit` option is provided to specify custom weights. Might be a simpler experience that way so Ill update with this in mind.

Yea I think that is a great idea. It will definitely be simpler.

This would also make this a breaking change which is good so we can get all the breaking changes in 1 release.

TarikGul (approved):
LGTM! 🚀

PR #410, opened 2024-05-24
chore(examples): update xmcVersion to safeXcmVersion

TarikGul:
This just ensures we no longer hardcode the xcmVersion but instead just pull from safeXcmVersion


marshacb (approved):
lgtm!

Issue #405, opened 2024-05-21
Improve/fix error messages


IkerAlus:
After adding features like bridges to the API certain error messages are not accurate. It would be nice to review the error cases and messages to make sure they are accurate and meaningful.

Some examples:

- When trying to transfer ROC back to Rococo relay from Rococo Asset Hub, the error message mentions Kusama:

`InvalidAsset: (SystemToRelay) assetId {"parents":"2","interior":{"X1":{"GlobalConsensus":"Westend"}}} not native to kusama. Expected KSM`

- When inputting an invalid XCM location, the message is outdated ("statemine") and does not consider XCM locations.

`InvalidAsset: assetId {"parents":"3","interior":{"X1":{"GlobalConsensus":"Westend"}}} is not a valid symbol or integer asset id for statemine`

PR #416, opened 2024-06-19
fix: improve error messages

marshacb:
This PR improves error messaging which previously only mentioned integer assetIDs and symbols as inputs by adding asset locations

closes: [#405](https://github.com/paritytech/asset-transfer-api/issues/405)


TarikGul (approved):
LGTM!

PR #444, opened 2024-10-08
fix: enable paysWithFeeDest for transferAssets

marshacb:
### Changes

This PR removes an XCM check which restricted the number of assets that could be provided when constructing a `transferAssets` extrinsic to 1. This allows for a second asset to be provided in the `assetIds` and `amounts` fields in order to be used with the `paysWithFeeDest` option for selecting a different asset to pay fees on the destination chain.


IkerAlus (approved):
LGTM

bee344 (approved):

PR #458, opened 2024-10-21
chore: declare explicit pjs dependencies

ryanleecode:
we import these PJS deps but they are transitive dependencies. they should be declared explictly in package.json. this will also allow dependabot to update them

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,12 @@
 		"typedoc-theme-hierarchy": "^4.0.0"
 	},
 	"dependencies": {
-		"@polkadot/api": "^14.0.1",
+		"@polkadot/api": "14.0.1",
+		"@polkadot/api-augment": "14.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,7 +50,12 @@
 		"typedoc-theme-hierarchy": "^4.0.0"
 	},
 	"dependencies": {
-		"@polkadot/api": "^14.0.1",
+		"@polkadot/api": "14.0.1",
```
TarikGul, line 53: Newest versions are 14.1.1, and 13.2.1

ryanleecode: Ill let dependabot update it. Trying to avoid changing the lockfile b.c i think updating will break the code

TarikGul: Makes sense, thanks!

TarikGul (approved):

PR #527, opened 2025-03-20
fix: Change package entry point

noplisu:
lib/index.js does not exist since version 0.4.5 while the package.json still points to it. The correct index.js is located in lib/src/index.js. This changes the package.json to point to the correct package entry point.


marshacb (approved):

marshacb:
@noplisu Thanks for the PR

Issue #544, opened 2025-04-22
getDestinationXcmWeightToFeeAsset throws an error with Polimec's xcmFeeAsset


Imod7:
### Description
When calling `getDestinationXcmWeightToFeeAsset` with `USDT` feeAsset and `Polimec` chain the following error is thrown:
```
InvalidPallet: unable to initialize xcAssets registry for polkadot
    at getAssetId (/Users/eric.shek/Developers/polkadot-poc/node_modules/@substrate/asset-transfer-api/src/createXcmTypes/util/getAssetId.ts:128:10)
    at Function.getDestinationXcmWeightToFeeAsset (/Users/.../Developers/polkadot-poc/node_modules/@substrate/asset-transfer-api/src/AssetTransferApi.ts:1458:38)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)

```

### Example Case
Example case to locally reproduce the error (with a locally linked package of asset-transfer-api): 
```
const { AssetTransferApi, constructApiPromise } = require('../asset-transfer-api'); // the locally linked package

const main = async () => {
    const { api, specName, safeXcmVersion } = await constructApiPromise('wss://polkadot-asset-hub-rpc.polkadot.io');

    const assetTransferApi = new AssetTransferApi(api, specName, safeXcmVersion);

    const sendersAddr = '16FNntR3F4ZjsTdYHysWSiP2vGHoVL7uZ9tyfyd4o5Cdaghf';
    const tx = '0x1f0b04010100c91f0400010100e823f55dbee3d394f448f2f74194078ba4259e4b31c28e155c29531106a8ea0804080002043205011f0002093d000100000700e40b54020000000000';
    const dryRunResult = await assetTransferApi.dryRunCall(sendersAddr, tx, 'call');
  
    const destinationFees = await AssetTransferApi.getDestinationXcmWeightToFeeAsset(
      'polimec-mainnet',
      'wss://rpc.polimec.org',
      4,
      dryRunResult,
      'USDT'
    ); 
    console.log('payload', JSON.stringify(destinationFees));
}
main();
```

### Root Cause
[...]

PR #546, opened 2025-04-24
fix: check if X1 exists in multilocation

Imod7:
### Description
Added this very small check because otherwise with : 
- this `asset-transfer-api-registry` update - PR [#180](https://github.com/paritytech/asset-transfer-api-registry/pull/180)
- and for this case in `asset-transfer-api` - Issue [#544](https://github.com/paritytech/asset-transfer-api/issues/544)

This line of code throws an error because it always expects for `xcmVersion > 3` to have `X1` in `result.interior` which I do not think it is the case. This can be double checked by using for example : 
- https://polkadot.js.org/apps/#/runtime connected to Polimec chain
- `xcmPaymentApi` - `queryAcceptablePaymentAssets`
- xcmVersion : 4
- the result will give also assets that have `interior.X3` and some other that have `interior.X1`


TarikGul (approved):
Nice job!

PR #562, opened 2025-05-17
chore: use tshy

ryanleecode:
See https://github.com/paritytech/asset-transfer-api/pull/561#pullrequestreview-2848336435

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Node Modules
-/node_modules
+node_modules
 
 # MacOS
 .DS_Store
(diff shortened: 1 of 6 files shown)


andrew-ifrita (approved):
This seems fine and is acceptable. I don't have a strong enough opinion to push back on this. Generally I prefer simplicity and more widely used tooling which led me to reach for `tsup` which has been around for a few more years (see: Lindy Effect) and has more than 10x the weekly downloads. It also has less config and also keeps package.json a bit cleaner. All leading to easier maintainability.

But alas, at the end of the day this accomplishes all the same and I am a relatively new contributor. If there are any issues, it is simple enough to revert.

Thanks for taking the time here to help me out. It was interesting comparing the two.

PR #565, opened 2025-05-20
feat: Add optional destApi argument to createTransferTransaction

andrew-ifrita:
Draft: In Progress

If provided, make additional calls to ensure that receiving accounts will be sufficient and that no assets will be lost when accounts are reaped.

This is somewhat of a band-aid, as accidental loss of assets is still possible at the protocol level. This at least improves the user experience when it comes to transferring assets until we can investigate fixing the root of the issue at the protocol level.

## General Idea

Throw and error or raise a warning **iff** all of the following are met

- tx is cross-chain (XCM)
- asset is not native
- asset is not sufficient
- destination account does not exist
    - destination account is below the existential deposit after tx would take place

## Motivation

Make it harder for assets to be accidentally be burned by sending them in insufficient accounts.

## Goal

- Don't break any existing API
- Verify receiving accounts are (or will be) sufficient
- Warn users

## Downsides

- Verification only happens at the time of the tx building
    - future account changes could still result in assets potentially being lost if the destinations are no longer sufficient
- Additional API calls

## Questions

- Do we want to throw and error and kill the tx building, or simply warn the user?
    - If we throw an error, should we add an optional bypass to ignore these checks


andrew-ifrita:
I can add a few tests, but it should be good enough for review at this point.

PR #581, opened 2025-06-05
fix: Example build

andrew-ifrita:
In progress

TODO:
- [ ] Update docs and README around building/running examples
- [ ] fix `build:examples` or simply remove it
- [ ] add CI step to make sure every example successfully runs

### DIFF
--- a/examples/WestendAssetHubToWestendCollectivesXcmV4.ts
+++ b/examples/WestendAssetHubToWestendCollectivesXcmV4.ts
@@ -3,9 +3,10 @@
  *
  * import { AssetTransferApi, constructApiPromise } from '@substrate/asset-transfer-api'
  */
-import { AssetTransferApi, constructApiPromise } from '../src';
-import { TxResult } from '../src/types';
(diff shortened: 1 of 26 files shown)


TarikGul (approved):
Overall looks good.

I noticed some of the examples are doing:

```typescript
void (async () => {
	try {
		await main();
	} catch (err) {
		console.error(err);
		process.exit(1);
	} finally {
		process.exit();
	}
})();
```

And others are the same as before, is there a reason for that?

andrew-ifrita:
> Overall looks good.
> 
> I noticed some of the examples are doing:
> 
> ```ts
> void (async () => {
> 	try {
> 		await main();
> 	} catch (err) {
> 		console.error(err);
> 		process.exit(1);
> 	} finally {
> 		process.exit();
> 	}
> })();
> ```
> 
> And others are the same as before, is there a reason for that?

@TarikGul Good catch. I will fix. Looks like my find/replace wasn't recursive.

PR #585, opened 2025-06-09
refactor: Remove isForeignAssetsTransfer from `fetchCallType`

andrew-ifrita:
It was redundant with assetType which is either Native or Foreign.


TarikGul (approved):
🚀

PR #590, opened 2025-06-11
feat: XCM v5

andrew-ifrita:
Adding support for XCM v5.

Refactoring in the process as there is quite a lot of duplication which made adding support more time intensive than need be.

Potential TODO (may be better for a later date):
- [ ] Consolidate XCM types to make it clear what differs between versions.


src/consts.ts
```diff
@@ -43,10 +43,10 @@ export const PASEO_ASSET_HUB_SPEC_NAME = ['asset-hub-paseo'];
  */
 export const DEFAULT_XCM_VERSION = 2;
 /**
- * There should only ever be three supported versions.
- * Therefore supported xcm versions will always have a fixed length of 3.
+ * Parity officially supports 3 versions at a time,
+ * but older chain could still accept older versions.
  */
-export const SUPPORTED_XCM_VERSIONS: [number, number, number] = [2, 3, 4];
+export const SUPPORTED_XCM_VERSIONS: number[] = [2, 3, 4, 5];
```
TarikGul, line 49: I think we can get rid of 2 here right? I don't think its still used anywhere.

TarikGul: Or we can keep it for backwards compatibility, not a big deal on this one.

andrew-ifrita: I actually started there with this simple PR https://github.com/paritytech/asset-transfer-api/pull/588 but then opted to get this in first.

Default should be 4 as that is currently marked as LTS. We will make sure to get that merged before the next release.

TarikGul (approved):
Overall, everything looks way better than previously. I looked over all the refactoring that DRY'ed the could up as well, and think its a big improvement, and more maintainable.

filvecchiato (approved):
Very good improvement!!

PR #596, opened 2025-06-18
refactor: Consolidate FungibleStrAssetType and FungibleObjAssetType

andrew-ifrita:
Consolidate `FungibleStrAssetType` and `FungibleObjAssetType` into `FungibleAssetType`

Both represent the same underlying structure.
There was an extra nested "Fungible" in ObjType as well that added unnecessary complexity.

edit: also consolidate `XcmAsset` and `XcmMultiAsset` into `FungibleAsset` and `FungibleMultiAsset`

## Review Notes

I don't think this is a Chesterton's Fence but can I get a slightly more thorough review on this one? There is no underlying Asset struct in `polkadot-sdk` with a nested `Fungible` like `FungibleObjAsset` had. I could not find any reasonable explanation for the separation of the Obj and Str types. All tests continue to pass after the changes.


TarikGul (approved):
Overall nice refactor, just one small suggestion.

That all being said, I was pretty thorough in PR and looking for the actual use of the Fungible Object in the SDK and polkadot-js. I couldn't find anything so I think its safe to say we can definitely remove it.

PR #598, opened 2025-06-20
refactor: Decouple XCM versioning from Direction logic handlers

andrew-ifrita:
The main goal here is the decoupling of the XCM version code from everything else. Some duplication was added but the net result should be positive. There is still a lot of room for improvement when it comes to cleaning everything up here.

## Motivation

Reorganize XCM version specific code so adding support for new versions is more straight forward.

Ideally only `src/createXcmTypes/xcm/v{version}.ts` and `src/consts.ts` should need to be touched (outside of tests).

## Key changes

1. Introduce `XcmCreator` interface that is shared by xcm versions
2. Add an `xcmCreator` to each logic handler for each direction (e.g. ParaToPara)
3. Separate xcm version formatting from transaction construction logic
4. add `src/createXcmTypes/handlers/` dir for the direction handlers 

## Naming

I'm not super happy with the naming I used. I suppose it is good enough but I am open to any suggestions for improvements here.

## Review Guide

It probably makes sense to view the changes roughly in this order

1. Structural and organizational changes
    - introduction of  `src/createXcmTypes/handlers/` and `src/createXcmTypes/xcm/`
2. High level interface changes in `types.ts`
3. `xcm/` dir in its entirety
4. How `xcmCreator` is implemented in the direction handlers (`handlers/`)
6. Tests and remaining changes

## TODO

- [x] Verify that I didn't leave any hanging TODOs in the comments / code

## Breaking Changes

- xcmVersion is no longer implicitly change to V4 when dealing with ForeignAssets transfers


src/AssetTransferApi.ts
```diff
@@ -406,10 +409,12 @@ export class AssetTransferApi {
 			},
 		);
 
+		const xcmCreator = getXcmCreator(declaredXcmVersion);
```
TarikGul, line 412: Just so can retierate and clarify the workflow here:

1. The xcmVersion is passed into getXcmCreator which returns back a versioned xcm api for constructing all parts of the xcm message.

2. The xcmCreator replaces anywhere the xcmVersion was threaded down functions.

3. Those downstream functions just use the xcm creator api that is passed down.

src/AssetTransferApi.ts
```diff
@@ -1261,8 +1269,9 @@ export class AssetTransferApi {
 				}
 				palletMethod = `poolAssets::${method}`;
 			} else if (localAssetType === LocalTxType.ForeignAssets) {
-				const foreignAssetsXcmVersion = 4;
-				const location = resolveMultiLocation(JSON.parse(assetId) as AnyJson, foreignAssetsXcmVersion);
+				const foreignAssetsXcmVersion = 4; // TODO: Clean up and don't imply xcm versions
+				const xcmCreator = getXcmCreator(foreignAssetsXcmVersion);
```
TarikGul, line 1273: We already create the xcmCreator above at the top of the function. Maybe here we can say if the version is not 4 or greater, throw an error, then just use the established xcmCreator.

src/AssetTransferApi.spec.ts
```diff
@@ -178,10 +181,12 @@ describe('AssetTransferAPI', () => {
 			});
 		});
 		it('Should construct the correct payload', async () => {
+			const xcmCreator = getXcmCreator(4);
```
TarikGul, line 184: This test is using versioned xcmCreator for 4, but the test seems to be testing for version 2.

src/util/resolveMultiLocation.spec.ts
```diff
@@ -1,29 +1,34 @@
 // Copyright 2023 Parity Technologies (UK) Ltd.
 
+import { getXcmCreator } from '../createXcmTypes/xcm';
 import { resolveMultiLocation } from './resolveMultiLocation';
 
 describe('resolveMultiLocation', () => {
 	it('Should correctly not throw an error when xcmVersion is 3 and the multiLocation does not contain a generalKey Junction', () => {
 		const str = `{"parents":0,"interior":{"here": null}}`;
-		const err = () => resolveMultiLocation(str, 3);
+		const err = () => resolveMultiLocation(str, xcmCreator);
```
TarikGul, line 9: Should we create `xcmCreator` before `err` here?

src/AssetTransferApi.ts
andrew-ifrita: That is correct. The pain of the previous style came about via

- version specific code spread over 25+ different files
- inconsistent use (first, second, fifth, seventh paremeter was xcmVersion)
- lack of hard version checks / inconsistent checks - i.e. `xcmVersion=7` actually results in V4 in many places
    - the existing version checks were also spread out

Otherwise it would have been fine, with a little bit of discipline, but this now forces (or strongly encourages) version specific code to be located together

src/AssetTransferApi.ts
andrew-ifrita: Also turned out to be a breaking change, but for the better. We had these tests 36c4b9362e passing because V2 was implicitly changed to V4

src/AssetTransferApi.spec.ts
andrew-ifrita: Good catch. I just went and removed the `xcmVersion` parameter entirely there so this mismatch can't happen at all. (57b7e413e488435)

src/AssetTransferApi.ts
TarikGul: > version specific code spread over 25+ different files
inconsistent use (first, second, fifth, seventh parameter was xcmVersion)
lack of hard version checks / inconsistent checks - i.e. xcmVersion=7 actually results in V4 in many places
the existing version checks were also spread out

Yea these are all problems, glad were finally sorting this out!

> Otherwise it would have been fine, with a little bit of discipline, but this now forces (or strongly encourages) version specific code to be located together.

From what I can see from this PR its going to be a whole lot easier to build support for any other XCM versions that come out.

src/createXcmTypes/handlers/default.ts
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	createFeeAssetItem(_api: ApiPromise, _opts: CreateFeeAssetItemOpts): Promise<number> {
+		throw new Error('Not Implemented');
+	}
+
+	// XTokens methods don't need to be included here
+	// I don't think it hurts to include them,
+	// BUT there may be code that checks if a method exists
+	// So be careful and double check this before merge.
```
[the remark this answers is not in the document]
andrew-ifrita: We have some cases like `transferMultiassets.ts` with
```ts
if (
		typeCreator.createXTokensAssets &&
		typeCreator.createXTokensFeeAssetItem &&
		typeCreator.createXTokensBeneficiary
	) {
```

but these explicitly throw an error if not caught so I think we are good here. Grepping for `createXTokens` didn't yield any other results.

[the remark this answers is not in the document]
andrew-ifrita: Comment removed in bef67a9ef

PR #601, opened 2025-06-26
docs: add top level NOTICE and remove individual copyright headers

andrew-ifrita:
Changes:
- Removing copyright header from each individual file
- Adding top level `NOTICE` with copyright information (see: https://www.apache.org/legal/src-headers.html#notice)

We had inconsistent use of the headers, so for the sake of consistency and simplicity, lets reduce everything to just the notice file. I have been told that this is sufficient

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,5 @@
+Asset Transfer API
+Copyright 2025 Parity Technologies Ltd.
+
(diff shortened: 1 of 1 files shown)


filvecchiato (approved):
LGTM!

PR #603, opened 2025-06-27
refactor: Consolidate some duplicated tests and remove a few redundant files

andrew-ifrita:
- The tests related to Default handler were duplicated for each direction handler. I also started using `test.each` to cut down on the boiler plate for each test for each version.

- `createXTokensFeeAssetItem` logic was moved into `default.ts` as it made no sense to live in a separate file in a utils dir.

- `resolveMultiLocation` file was removed as it simply called `xcmCreator.resolveMultiLocation`

### DIFF
--- a/src/util/resolveMultiLocation.spec.ts
+++ b/src/util/resolveMultiLocation.spec.ts
@@ -1,11 +1,10 @@
 import { getXcmCreator } from '../createXcmTypes/xcm';
-import { resolveMultiLocation } from './resolveMultiLocation';
 
 describe('resolveMultiLocation', () => {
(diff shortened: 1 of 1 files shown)


andrew-ifrita:
This failing test

```
AssetTransferApi Integration Tests > getDestinationXcmWeightToFeeAsset > Correctly throws an error when the provided runtime does not support the xcmPaymentApi
```

Is unrelated to these changes and is also happening on master

TarikGul (approved):

filvecchiato (approved):

andrew-ifrita:
> Some of these integration tests that are flakey I am totally cool with just removing.

PR #635, opened 2025-07-24
fix: Check assets individually in `claimAssets(...)`

andrew-ifrita:
This is in place of treating all assets as the same if a single one contains a multilocation.

Also used this as an opportunity to move `claimAssets()` to object params as well as remove `CreateXcmCallOpts` parameter from `claimAssets()` as it was out of place.


TarikGul (approved):
👍

PR #641, opened 2025-07-30
feat: Add runtime validation for XcmMultiLocation

andrew-ifrita:
Allow throwing error on invalid input while parsing multilocations.

Motivation: I'm trying to debug some issues cause from attempting to use assetIds as MultiLocations, leading to unexpected behavior. This will correctly catch these.

Alternatively we could use something like `zod` but I was trying to keep this lightweight. If we need any future runtime validation, that is the route that we should go.

### DIFF
--- a/src/createXcmTypes/xcm/common.ts
+++ b/src/createXcmTypes/xcm/common.ts
@@ -1,6 +1,7 @@
 import { AnyJson } from '@polkadot/types-codec/types';
 import { isEthereumAddress } from '@polkadot/util-crypto';
 
+import { SUPPORTED_XCM_VERSIONS } from '../../consts.js';
 import { BaseError, BaseErrorsEnum } from '../../errors/BaseError.js';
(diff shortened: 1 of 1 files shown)


src/createXcmTypes/xcm/common.ts
```diff
@@ -64,3 +66,43 @@ export function parseMultiLocation(input: AnyJson, xcmVersion: number): XcmMulti
...
+ * This is not exhaustive and is a bit messy.
+ * The output may not actually be a XcmMultiLocation but
+ * a MultiLocationVariant<J> instead. This dates back to
+ * some historical string parsing.
+ *
+ * Ideally this is eventually removed altogether with the
+ * removal of passing stringified multilocations around
+ * to be interpretted.
+ */
+function validateXcmMultiLocation(obj: AnyJson): obj is XcmMultiLocation {
```
TarikGul, line 82: Yea this makes sense, a bit hacky but I think the merit is there!

TarikGul (approved):

PR #647, opened 2025-08-11
feat: Verify originChainId matches the chain id looked up from specName

### LINKED ISSUE #625: `originChainId` is derived from the given specName
The `originChainId` should instead be derived from the api or at least confirmed against the api. If not, human error can lead to some harder to diagnose bugs due to the wrong chain id being inferred which can in turn impact the transaction type.

andrew-ifrita:
Resolves #625 

Verify that the `originChainId` matches the parachain id of the supplied API, if not, throw an error. This is meant to catch human error where the wrong spec name is supplied which can lead to hard to debug issues.

## Testing

Testing of these changes is covered via the existing [integration tests](https://github.com/paritytech/asset-transfer-api/actions/runs/16882167519/job/47820514240?pr=647#step:7:1) and [examples](https://github.com/paritytech/asset-transfer-api/actions/runs/16882167519/job/47820514229?pr=647#step:7:27) which call `createTransferTransaction`. The unit tests are less useful here as the API responses have been mocked.

### DIFF
--- a/src/AssetTransferApi.ts
+++ b/src/AssetTransferApi.ts
@@ -211,6 +211,8 @@ export class AssetTransferApi {
 			xcmPalletOverride,
 		} = opts;
 
+		await this.verifyOriginChainId();
+
 		const { api, specName, safeXcmVersion, originChainId, registry } = this;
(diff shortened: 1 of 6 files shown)


filvecchiato (approved):
Looking good!

